PR #24, opened 2015-02-13
Add support for async generate

ferhatb:
Generators will likely need to read a template referenced from the annotated class.
Since all I/O is async, changed generate signature and GeneratedOutput.output to futures.


kevmoo:
I ran with my own implementation here - 06379e728180c9c8c19373f0c249c402e905e05b

Thanks!

PR #36, opened 2015-02-20
Fix matchAnnotation() function

### LINKED ISSUE #35: json_serializable_generator can't find serializable classes
I've created the following dart package to test source_gen:

[https://github.com/luizmineo/source_gen_test](https://github.com/luizmineo/source_gen_test)

It contains a `build.dart` script, which is invoked by DartEditor, and the `lib/example_json.dart` library, which has a serializable class. When modifying this file, `lib/example_json.g.dart` should be created, but it seems that source_gen can't find the `Person` class in example_json.dart, and nothing happens.

luizmineo:
Fix #35

### DIFF
--- a/lib/src/annotation.dart
+++ b/lib/src/annotation.dart
@@ -126,10 +126,10 @@ bool matchAnnotation(Type annotationType, ElementAnnotationImpl annotation) {
 
   switch (libOwner.uri.scheme) {
     case 'file':
-      libraryUri = libOwner.uri;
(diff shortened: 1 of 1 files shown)


luizmineo:
The `_fileUriFromPackageUri()` has an assert that checks the scheme of the source library for a specific value, but it works fine when invoked by DartEditor.

Now I'm not sure if this is really a bug, or if I'm not doing it right...

kevmoo:
Good try. I need to approach this another way. Stay tuned.

PR #51, opened 2015-06-03
Match annotations defined in parts, add relevant unit tests

greglittlefield-wf:
#### Ultimate Problem
Annotations defined in parts of libraries are not matched by generators, which is unexpected behavior. (Annotations shouldn't always have to be defined in a main library file.)

#### Changes
* Match annotations based on equality between owner library URI and source _library_ uri, as opposed to owner library uri and source _file_ URI.
* Update tests.


sethladd:
cc @kevmoo

greglittlefield-wf:
My company, Workiva, signed the CLA.

PR #64, opened 2015-09-25
Add null aware operator before calling toList.

tayloranderson-wf:
If a list type is null, the call to map() is prevented, but toList() is still called resulting in a NPE.


tayloranderson-wf:
I signed it!

kevmoo:
Thanks, @tayloranderson-wf 

Help me in the future: feel free to update the changelog. Also: regression tests are always appreciated. See the commits I did before/after I merged your patch in.

Cheers.

PR #66, opened 2015-11-11
Add "jsonKey" annotation

### LINKED ISSUE #65: Allow customization of the resulting JSON keys
In Go, the serialized JSON for a struct can be customized using this syntax:

```go
type Foo struct {
	BarBaz string `json:"bar"`
	...
}
```

It would be great if there was a similiar annotation for properties, so that classes could use a more meaningful names for their Dart API:

```dart
@JsonSerializable()
class Foo extends Object with _$FooSerializerMixin {
  @Json("bar")
  final String barBaz;
  ...
}
```

resulting in 

```json
{
    "bar": "_"
}
```

johnpryan:
This fixes #65.
This adds the ability to specify a `JsonKey` annotation:
```dart
@JsonSerializable()
class MyClass {
  @JsonKey("foo")
  final int FooBar;
  ...
```

source_gen then uses that value for the key when serializing to JSON: 

```json
{
    "foo": 123
}
```

FYI @kevmoo

### DIFF
--- a/lib/generators/json_serializable.dart
+++ b/lib/generators/json_serializable.dart
@@ -15,3 +15,12 @@ class JsonSerializable {
       : this.createFactory = createFactory,
         this.createToJson = createToJson;
 }
+
+/// Customizes the name of the JSON key for a field.
+///
+/// If ommitted, the resulting JSON key will be the
+/// name of the field defined on the class.
+class JsonKey {
+  final String jsonName;
+  const JsonKey(this.jsonName);
+}
--- a/test/json_serializable_test.dart
+++ b/test/json_serializable_test.dart
@@ -100,6 +100,14 @@ void main() {
       expect(output, contains('json[\'children\']?.map('));
     });
   });
+
+  test('reads JsonKey annotations', () async {
+    var element = await _getClassForCodeString('Person');
+    var output = await _generator.generate(element);
+
+    expect(output, contains("'h': height,"));
+    expect(output, contains("..height = json['h']"));
+  });
 }
 
 const _generator = const JsonSerializableGenerator();
@@ -144,6 +152,7 @@ void annotatedMethod() => null;
 @JsonSerializable()
 class Person {
   String firstName, lastName;
+  @JsonKey("h")
   int height;
   DateTime dateOfBirth;
   dynamic dynamicType;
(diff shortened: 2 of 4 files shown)


johnpryan:
I signed it!

johnpryan:
it looks like the CI tests are failing on master too

kevmoo:
Gah! Looks like we have an analyzer breaking change. We may need to fix this before you change goes in...hmm...

kevmoo:
@johnpryan it seems that we have breaking changes in analyzer since `0.26.0`. Investigating

kevmoo:
See https://github.com/dart-lang/source_gen/issues/67

kevmoo:
@johnpryan try rebasing on master – I just set an upper bound on the analyzer version while we track down the problem

johnpryan:
@kevmoo done - CI is passing now

travissanderson-wf:
very nice!!

test/json_serializable_test.dart
```diff
@@ -100,6 +100,16 @@ void main() {
       expect(output, contains('json[\'children\']?.map('));
     });
   });
+
+  group('reads Json field annotations', () {
+    test('1', () async {
```
kevmoo: just make this a test w/out a group – if there is only one test

lib/generators/json_serializable.dart
```diff
@@ -15,3 +15,8 @@ class JsonSerializable {
       : this.createFactory = createFactory,
         this.createToJson = createToJson;
 }
+
+class JsonProperty {
```
kevmoo: Be better than me: add some doc comments to this class. :smile:

johnpryan: done

test/json_serializable_test.dart
johnpryan: done

johnpryan:
@kevmoo ready for another look

johnpryan:
@kevmoo This is ready for another look.  Thanks!

kevmoo:
SO CLOSE, @johnpryan – update `CHANGELOG.md` and we're good to go

johnpryan:
@kevmoo done!

I used `0.4.7` for the tag number in `CHANGELING.md` since it looks like this repo uses the 'patch' number for backwards compatible API changes.

kevmoo:
Landed w/ tweaks here: https://github.com/dart-lang/source_gen/commit/e32619ef247f09020bec64fe8ff2da1077dcbf05

Thank you!

PR #68, opened 2015-11-19
support asset paths for annotation library sources

jakemac53:


pubspec.yaml
```diff
@@ -12,6 +12,7 @@ dependencies:
   args: '>=0.12.2+3 <0.14.0'
   cli_util: ^0.0.1
   dart_style: '>=0.1.7 <0.3.0'
+  mockito: ^0.11.0
```
kevmoo: make this a dev_dependency

test/annotation_test.dart
```diff
@@ -221,12 +224,25 @@ void main() {
...
+        test('When running on a normal file system', () {
+          ClassElement annotatedClass =
+              _getAnnotationForClass(libElement, 'AnnotatedWithJson');
+          var annotation = annotatedClass.metadata.single;
+          var matched = matchAnnotation(JsonSerializable, annotation);
+          expect(matched, isTrue);
+        });
+
+        test('When running in barback', () {
+          var annotation = _mockElementAnnotation(
```
kevmoo: Is there anyway to do this without mock silly? I worry about this being brittle...

jakemac53: Ya, its definitely brittle. All the other ways I can think of to test this are a lot worse though, and would require a barback dependency and mocking out a bunch of barback stuff :/

kevmoo:
LGTM

PR #77, opened 2016-01-30
Use pkg/resource to find paths

kevmoo:
The async-ness of the API hurts a bit.

But it feels less broken than using mirrors.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: source_gen
-version: 0.4.8
+version: 0.4.9-dev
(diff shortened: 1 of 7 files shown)


jakemac53:
I don't think there is really that much of an issue with using mirrors for this use case? The additional setup step this adds is worse IMO than the mirrors hacks.

jakemac53:
Alternatively you could just change getPackagePath() to be async

kevmoo:
Yeah, then you leak async everywhere.

I'll kill this off. Just getting a feel for the new API.

PR #80, opened 2016-03-10
Use build pkg

jakemac53:


kevmoo:
@jakemac53 I pushed another commit w/ a couple of tweaks. Review. Rebase. Land it!

jakemac53:
Dev bots are failing because test didn't support the dev sdk until very recently. Going to wait a bit and try re-running to make sure everything is ok.

jakemac53:
Actually it looks like scheduled_test needs to be updated to support the latest test

jakemac53:
Ok, I actually just removed scheduled_test it entirely as a dep since it was no longer being used. I also moved `io.dart` under the test folder since that was no longer used by the main libs.

PR #82, opened 2016-04-28
Support latest version of build, upgrade analyzer dev

kevmoo:
Also stop using most deprecated analyzer APIs

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -8,19 +8,19 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:analyzer/analyzer.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/file_system/file_system.dart' hide File;
 import 'package:analyzer/file_system/physical_file_system.dart';
 import 'package:analyzer/source/package_map_provider.dart';
 import 'package:analyzer/source/package_map_resolver.dart';
 import 'package:analyzer/source/pub_package_map_provider.dart';
-import 'package:analyzer/src/generated/ast.dart';
-import 'package:analyzer/src/generated/element.dart';
 import 'package:analyzer/src/generated/engine.dart';
 import 'package:analyzer/src/generated/java_io.dart';
-import 'package:analyzer/src/generated/sdk_io.dart' show DirectoryBasedDartSdk;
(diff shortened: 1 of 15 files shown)


lib/src/utils.dart
```diff
@@ -8,19 +8,19 @@ import 'dart:async';
...
 import 'package:analyzer/source/package_map_resolver.dart';
 import 'package:analyzer/source/pub_package_map_provider.dart';
-import 'package:analyzer/src/generated/ast.dart';
-import 'package:analyzer/src/generated/element.dart';
 import 'package:analyzer/src/generated/engine.dart';
 import 'package:analyzer/src/generated/java_io.dart';
-import 'package:analyzer/src/generated/sdk_io.dart' show DirectoryBasedDartSdk;
 import 'package:analyzer/src/generated/sdk.dart' show DartSdk;
-import 'package:analyzer/src/generated/source_io.dart';
+import 'package:analyzer/src/generated/sdk_io.dart' show DirectoryBasedDartSdk;
```
jakemac53: This (and the entire `getAnalysisContextForProjectPath` method) shouldn't actually be necessary any more? I must have missed deleting that when I switched to pkg/build.

jakemac53: Actually now that I think about it I think particle is using this though... and it looks like this package is using it in tests.

jakemac53:
lgtm, may want to move `getAnalysisContextForProjectPath` to the test folder or something, but since it is being relied on externally maybe not.

pubspec.yaml
```diff
@@ -1,13 +1,13 @@
...
 author: Dart Team <misc@dartlang.org>
 description: Automatic sourcecode generation for Dart
 homepage: https://github.com/dart-lang/source_gen
 environment:
   sdk: '>=1.12.0 <2.0.0'
 dependencies:
-  analyzer: '^0.27.3-alpha.1'
-  build: ^0.2.1
+  analyzer: '^0.27.3'
+  build: '>=0.2.1 <0.4.0'
```
jakemac53: Technically the tests wouldn't pass in 0.2.x, but I think thats ok. It does actually support that version, its just the enum values which appear in the test that would be broken.

kevmoo: Exactly 😄

lib/src/utils.dart
kevmoo: ...so leave it as is?

PR #84, opened 2016-06-06
Fix multiline error output for more than two lines of error message.

davidmorgan:
Currently lines after the second one are concatenated without newlines, leading to error output like:

// Error: first line of error message.
//        Second line of error message. //        Third line of error message.        // Fourth line of error message.


jakemac53:
lgtm

kevmoo:
published as 0.5.0+3 – if you want to bring it into google3, etc

PR #86, opened 2016-07-25
Added field isStandalone to GeneratorBuilder to allow generation of a…

samschendel:
… file that is not part of the source file.

### DIFF
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -14,8 +14,10 @@ import 'utils.dart';
 class GeneratorBuilder extends Builder {
   final List<Generator> generators;
   final String generatedExtension;
+  final bool isStandalone;
 
-  GeneratorBuilder(this.generators, {this.generatedExtension: '.g.dart'}) {
+  GeneratorBuilder(this.generators,
+      {this.generatedExtension: '.g.dart', this.isStandalone: false}) {
     // TODO: validate that generatedExtension starts with a `.'
     //       not null, empty, etc
   }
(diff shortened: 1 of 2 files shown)


kevmoo:
Tests?

samschendel:
Added two tests for this option just a moment ago.  Let me know what you think.

kevmoo:
@jakemac53 thoughts on this?

jakemac53:
How would you add imports in this scenario? Previously since it was a part you would just have access to  everything the original library has access to (and could add imports there if needed as well).

In general I am not opposed but I think we need to add some mechanism for adding imports as well.

natebosch:
The mechanism for adding imports would be the same as the mechanism for adding code - we'd write it to the output.

We have some code built on top of this that works great with that approach.

jakemac53:
The issue with that is the generator doesn't have to much control specifically over where in the file its output goes. You can return code with some imports and if you are the only generator or the first generator that will work. However if somebody added another generator before yours then you would get syntax errors.

That might not be a strict blocker, but it does make standalone generators a bit more fragile.

lib/src/builder.dart
```diff
@@ -20,6 +20,10 @@ class GeneratorBuilder extends Builder {
       {this.generatedExtension: '.g.dart', this.isStandalone: false}) {
     // TODO: validate that generatedExtension starts with a `.'
     //       not null, empty, etc
+    if (this.generators.length > 1) {
```
jakemac53: && isStandalone

samschendel:
ArgumentError is now thrown when one attempts to create a GeneratorBuilder with more than one generator and has set isStandalone to true.  Added two tests to verify that this error is thrown under that circumstance but not thrown when two generators are used and isStandalone is false.

jakemac53:
lgtm, have you ran the formatter?

samschendel:
Did just now.

jodinathan:
Hi,

Is there an example of creating a standalone file for a group of files?
All examples I found are one generated file per parsed file.

natebosch:
See some discussion here:
https://github.com/dart-lang/build/issues/213

package:build, and therefore package:source_gen, have a restriction of 1 input -> 1 or more outputs whether the outputs can only change the file extension.

One workaround available today is to create a placeholder file and use that as the input, then use the glob api to find the group of files that are interesting and use them to generate the output.

PR #89, opened 2016-09-07
Fix NPE when ElementAnnotation.constantValue returns null.

alorenzen:
Instead of throwing a NPE, throw an ArgumentError with a message to suggest that a dependency might be missing.


natebosch:
LGTM. Thanks!

PR #91, opened 2016-09-13
avoid computeNode calls

natebosch:
This makes annotation.dart forward compatible with implementations of the Resolvers and LibraryElement using analyzer summary files.

### DIFF
--- a/lib/src/annotation.dart
+++ b/lib/src/annotation.dart
@@ -177,7 +177,7 @@ dynamic _createFromConstructor(
     }
   }
 
-  Symbol ctorName = new Symbol(ctor.name ?? '');
+  var ctorName = new Symbol(ctor.name ?? '');
 
   var declarationMirror =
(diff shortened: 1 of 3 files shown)


sigmundch:
nice! lgtm

lib/src/annotation.dart
```diff
@@ -178,12 +177,7 @@ dynamic _createFromConstructor(
     }
   }
 
-  Symbol ctorName;
-  if (ctorDeclaration.name == null) {
-    ctorName = const Symbol('');
-  } else {
-    ctorName = new Symbol(ctorDeclaration.name.name);
-  }
+  Symbol ctorName = new Symbol(ctor.name ?? '');
```
kevmoo: use `var ctorName =`

natebosch: Good call. Done

PR #98, opened 2016-11-04
Add an incremental generator

### LINKED ISSUE #33: Support incremental generations
I like to be able to have a source like

``` dart
@Value()
abstract class _MyClass {
  String a, b;
}
```

The generator for `ValueClass` whould generate:

``` dart
@DefaultConstructor()
@EqualsAndHashCode()
@ToString()
@ToConcretPublicClass()
abstract class _MyClass {
  String a, b;
}
```

Then the generated part would be re-analyzed and the generator for `DefaultConstructor` would generate:

``` dart
@EqualsAndHashCode()
@ToString()
@ToConcretPublicClass()
abstract class _MyClass {
  String a, b;
  _MyClass({this.a, this.b});
}
```

Next round with  the generator for `EqualsAndHashCode`:

``` dart
@ToString()
@ToConcretPublicClass()
abstract class _MyClass {
  String a, b;
  _MyClass({this.a, this.b});
  @override int get hashCode => hashObjects([a, b]);
  @override bool operator ==(o) => identical(this, o) || o.runtimeType == runtimeType && o.a == a && o.b == b;
}
```

Next round with  the generator for `ToString`:

``` dart
@ToConcretPublicClass()
abstract class _MyClass {
  String a, b;
  _MyClass({this.a, this.b});
  @override int get hashCode => hashObjects([a, b]);
  @override bool operator ==(o) => identical(this, o) || o.runtimeType == runtimeType && o.a == a && o.b == b;
  @override String toString() => "_MyClass(a=$a, b=$b)";
}
```

Next round with  the generator for `ToConcretPublicClass`:

``` dart
class MyClass {
  String a, b;
  MyClass({this.a, this.b});
  @override int get hashCode => hashObjects([a, b]);
  @override bool operator ==(o) => identical(this, o) || o.runtimeType == runtimeType && o.a == a && o.b == b;
  @override String toString() => "MyClass(a=$a, b=$b)";
}
```

This issue is a little related to #28

a14n:
Fixes #33

### DIFF
--- a/lib/src/incremental_generator.dart
+++ b/lib/src/incremental_generator.dart
@@ -25,9 +25,11 @@ abstract class IncrementalGenerator extends Generator {
   Future<String> generate(Element element, BuildStep buildStep) async {
     if (element is! LibraryElement) return null;
     final lib = element as LibraryElement;
+    final generatedPart = getGeneratedPart(lib);
+
+    if (generatedPart == null) return null;
 
     // back up the initial content of the part to restore at the end
-    final generatedPart = getGeneratedPart(lib);
     String genContent, initialContent;
     try {
       initialContent = lib.context.getContents(generatedPart.source).data;
@@ -68,7 +70,8 @@ abstract class IncrementalGenerator extends Generator {
     final genPartName = _getGeneratedPartName(lib);
     final genPartPath = path
         .normalize(path.join(path.dirname(lib.source.uri.path), genPartName));
-    return lib.units.firstWhere((u) => u.source.uri.path == genPartPath);
+    return lib.units.firstWhere((u) => u.source.uri.path == genPartPath,
+        orElse: () => null);
   }
 
   /// Returns the file name of the generated part
--- a/lib/src/incremental_generator.dart
+++ b/lib/src/incremental_generator.dart
@@ -1,0 +1,79 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/src/generated/engine.dart';
+import 'package:build/build.dart';
+import 'package:path/path.dart' as path;
+
+import 'generator.dart';
+
+/// A generator that will try to generate recursively until 2 consecutive
+/// generations output the same result.
+abstract class IncrementalGenerator extends Generator {
+  final int maxIterations;
+
+  const IncrementalGenerator({this.maxIterations: 1000});
+
+  Future<String> generateForLibraryElement(
+      LibraryElement library, BuildStep buildStep);
+
+  @override
+  Future<String> generate(Element element, BuildStep buildStep) async {
+    if (element is! LibraryElement) return null;
+    final lib = element as LibraryElement;
+
+    // back up the initial content of the part to restore at the end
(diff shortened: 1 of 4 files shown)


lib/src/incremental_generator.dart
```diff
@@ -0,0 +1,79 @@
...
+
+      final nextGenContent = await generateForLibraryElement(lib, buildStep);
+
+      // exit if stable
+      if (nextGenContent == genContent) break;
+
+      genContent = nextGenContent;
+
+      // next increment : add current genContent to initial content
+      lib.context.applyChanges(
```
jakemac53: Applying changes to the analyzers view of the world is going to be problematic. As a performance measure we use a single `Resolvers` instance for everything, so this will affect all generators view of the world. This also makes these files available to other builders in the same phase, when it isn't supposed to be available until the next phase. This could lead to some very subtle issues :(.

Also I don't know how this will behave with multiple generators in the same file, I don't think it will end well.

lib/src/incremental_generator.dart
```diff
@@ -0,0 +1,79 @@
...
+          new ChangeSet()..changedContent(generatedPart.source, genContent));
+    }
+
+    // reset part to its initial content
+    lib.context.applyChanges(initialContent == null
```
jakemac53: Ah I presume this is designed to mitigate my previous comment, but unfortunately I don't think it really solves the issue due to the async nature of this method. Any other generator in the same phase could be running interleaved with this one (this is part of why its important that we don't expose this generated file to other generators in the same phase, so they all have a uniform view of the world).

lib/src/incremental_generator.dart
```diff
@@ -0,0 +1,79 @@
...
+    final genPartName = _getGeneratedPartName(lib);
+    final genPartPath = path
+        .normalize(path.join(path.dirname(lib.source.uri.path), genPartName));
+    return lib.units.firstWhere((u) => u.source.uri.path == genPartPath);
+  }
+
+  /// Returns the file name of the generated part
+  String _getGeneratedPartName(LibraryElement lib) {
+    final name = lib.source.shortName;
+    return name.substring(0, name.length - '.dart'.length) + '.g.dart';
```
jakemac53: the `path` package has a `changeExtension` function which would be nice here ;)

a14n:
> Just to clarify, you want to do it like this so that you can analyze the output of another generator and update the generated output based on that?

Exactly. The analyzer have to provide an up to date structure at every roundtrip. My initial use case is #33.

lib/src/incremental_generator.dart
a14n: I don't find `changeExtension` in [path-1.4.0](https://www.dartdocs.org/documentation/path/1.4.0/).

jakemac53: heh, I guess I just made that up :). There is `withoutExtension` though which you could use to get rid of the `.dart`.

a14n: Yep I saw and used it :)

lib/src/incremental_generator.dart
a14n: I tried to implement a solution based on cloning of the library and its generated part but I don't really succeed to add them without errors (the resolver do not resolve the cloned part).
Actually I don't like this cloning solution.
Finally I wonder if we couldn't introduce a specific kind of generator to ensure that if this kind of generator is used then this generator must be the only one generator for a Phase. What do you think ?

a14n:
Closing as it doesn't seem to satisfy the requirements. I really hope that #33 will be eventually fixed in an other way.

PR #103, opened 2016-11-16
Use pkg/build_test, delete some duplicate logic

kevmoo:
A good start – need to figure out what's needed to use `testPhases` from `build_test`


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: source_gen
-version: 0.5.1+6
+version: 0.5.2-dev
```
natebosch: whats your reason for choosing .2 instead of +7?

PR #106, opened 2016-11-18
fix #104 generate valid lists in strong mode

### LINKED ISSUE #104: JsonSerializable: generate valid lists in strong mode
Currently, lists will generate a constructor like this:

```dart
// source
class User extends Object with _$UserSerializerMixin {
  List<int> tags;
  User(this.tags);
  factory User.fromJson(Map json) => _$UserFromJson(json);
}

// generated
User _$UserFromJson(Map json) => new User(
(json['tags'] as List)?.map((v0) => v0)?.toList());
```

Which generates the following strong-mode error:

```
WARNING: Unsafe implicit cast from 'List<dynamic>' to 'List<int>'. This usually indicates that type information was lost and resulted in 'dynamic' and/or a place that will have a failure at runtime. ([revolution] lib/model.g.dart:725)
```

a14n:


jakemac53:
By default strong mode would insert implicit casts for these afaik, do you have that disabled?

a14n:
See #104

jakemac53:
Ah, nvm I see in the latest sdk that the `List<int>` case is broken.

a14n:
btw should the generated code be `strong-mode: true` error-free or `strong-mode: implicit-casts: false` error-free ? My PR puts the generated code between the two setup :)

PR #111, opened 2016-11-30
Fix check for implemented interfaces

natebosch:
The `interfaces` getter does not behave as expected. See
https://github.com/dart-lang/sdk/issues/27947

This code was safe before since List happened to explicitly have
`implements Iterable`, but now that there is an
`EfficientLengthIterable` interface in between it isn't working.

Recursively check interfaces the same way we recursively check the
superclass. This now ignores the edge case where multiple interfaces
matched the tester and will return the first match.


natebosch:
Added changelog. Version was bumped in the last commit so I'll publish with this fix

PR #116, opened 2017-01-05
Extract matchTypes from matchAnnotation.

alorenzen:
Extract matchTypes from matchAnnotation so that it can be used in cases where the type comes from a different source (like a DartObject instance).

### DIFF
--- a/lib/src/annotation.dart
+++ b/lib/src/annotation.dart
@@ -214,6 +214,8 @@ DeclarationMirror _getDeclarationMirrorFromType(InterfaceType type) {
   return libMirror.declarations[typeNameSymbol];
 }
 
+/// Checks whether the constant type of [annotation] is equivalent to
+/// [annotationType].
 bool matchAnnotation(Type annotationType, ElementAnnotation annotation) {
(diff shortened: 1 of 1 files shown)


lib/src/annotation.dart
```diff
@@ -215,14 +215,18 @@ DeclarationMirror _getDeclarationMirrorFromType(InterfaceType type) {
...
   var annotationValueType = annotation.constantValue?.type;
   if (annotationValueType == null) {
     throw new ArgumentError.value(annotation, 'annotation',
         'Could not determine type of annotation. Are you missing a dependency?');
   }
+  
+  return matchTypes(annotationType, annotationValueType);
+}
+
+bool matchTypes(Type annotationType, ParameterizedType annotationValueType) {
```
natebosch: can you add a Doc comment?

natebosch:
I'm on board with the change. Will wait to merge it in until I get a chance to sync with @kevmoo

lib/src/annotation.dart
alorenzen: Done

PR #118, opened 2017-01-07
Drop unused methods from utils.dart

natebosch:
This file is not exported so there (hopefully) shouldn't be usage of any
of these methods outside of this package.

- Drop the methods that have no usage in this package
- Move the methods used only from tests to a library under test/


natebosch:
We could probably do without some of these methods even in the tests... Moving it out of lib/ is the low hanging fruit

PR #121, opened 2017-01-09
Make use of analyzer's new resolutionMap interface.

stereotype441:
The getters in the analyzer's AST data structures will soon be changed to less specific types, so we need to use the new interface in order to avoid warnings.

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/standard_resolution_map.dart';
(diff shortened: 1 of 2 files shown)


lib/src/utils.dart
```diff
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/standard_resolution_map.dart';
```
kevmoo: Does this effectively bump the minimum pkg/analyzer version?

Any pubspec changes required?

stereotype441: Whoops, you're right.  I forgot to update that.  Fixing...

kevmoo: Cool!

Issue #115, opened 2016-12-13
The builder should ignore 'part of' files


natebosch:
If `bar.dart` is a `part of` another file it doesn't make sense to create `bar.g.dart`. I think we can recurse into libraries and treat every Element in `bar.dart` as if it was in the library `foo.dart`

alorenzen:
This is blocking https://github.com/dart-lang/angular2/issues/48

PR #123, opened 2017-01-10
Smooth degradataion for non-library assets

natebosch:
If we can't get a LibraryElement for a primary input it should not have
any output. Return early rather than throw later.


natebosch:
This will partially resolve https://github.com/dart-lang/source_gen/issues/115

We'll also need a fix in Resolver from package:build and build_barback

PR #125, opened 2017-01-11
Always release the Resolver instance

natebosch:


kevmoo:
Why? ...just curious...

natebosch:
It's required that we always release a Resolver or things blow up in code_transformers.
Things happen to be blowing up in a particularly nasty way where we can't catch the exception no matter what...

I filed https://github.com/dart-lang/build/issues/162 to see if we can make this easier.

Issue #130, opened 2017-02-09
Output failing file or build step on dart format failure


matanlurey:
> ```
> [SEVERE]: Error formatting the generated source code.
> This may indicate an issue in the generated code or in the formatter.
> Please check the generated code and file an issue on source_gen
> if approppriate.
> Could not format because the source could not be parsed:
> 
> line 1, column 6: Part-of directives must include a library name.
> part of ;
> ```

I don't have any information on what the failure is related to.

kevmoo:
Help us with the code that was run?

matanlurey:
FWIW, I figured it out - I was running a `Generator` that was outputting a `.json` file, but I forgot to say it was `isStandlone: true` - adding this line fixed it. I think if you don't have this set it assumes it is a Dart file and tries running the formatter, failing.

PR #132, opened 2017-02-13
Add more context about dartfmt failures

matanlurey:
Closes https://github.com/dart-lang/source_gen/issues/130.

I validated internally this produces a nicer error message, such as:

```
[SEVERE]: Error formatting the generated source code for <REDACTED>|lib/focus_trap.dart which was output to lib/focus_trap.template.dart.
This may indicate an issue in the generated code or in the formatter.
Please check the generated code and file an issue on source_gen if appropriate.
Could not format because the source could not be parsed:

<REDACTED>
```

### DIFF
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -78,10 +78,13 @@ class GeneratorBuilder extends Builder {
       genPartContent = formatter.format(genPartContent);
     } catch (e, stack) {
       buildStep.logger.severe(
-          """Error formatting the generated source code.
-This may indicate an issue in the generated code or in the formatter.
-Please check the generated code and file an issue on source_gen
-if approppriate.""",
+          ''
+          'Error formatting generated source code for ${library.identifier} '
+          'which was output to ${_generatedFile(buildStep.inputId).path}.\n'
+          'This may indicate an issue in the generated code or in the formatter.'
+          '\n'
+          'Please check the generated code and file an issue on source_gen if '
+          'appropriate.',
           e,
           stack);
     }
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -78,13 +78,10 @@ class GeneratorBuilder extends Builder {
       genPartContent = formatter.format(genPartContent);
     } catch (e, stack) {
       buildStep.logger.severe(
-          ''
-          'Error formatting generated source code for ${library.identifier} '
-          'which was output to ${_generatedFile(buildStep.inputId).path}.\n'
-          'This may indicate an issue in the generated code or in the formatter.'
-          '\n'
-          'Please check the generated code and file an issue on source_gen if '
-          'appropriate.',
+          '''Error formatting generated source code for ${library.identifier}
+which was output to ${_generatedFile(buildStep.inputId).path}.\n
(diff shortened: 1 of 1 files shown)


lib/src/builder.dart
```diff
@@ -78,10 +78,13 @@ class GeneratorBuilder extends Builder {
       genPartContent = formatter.format(genPartContent);
     } catch (e, stack) {
       buildStep.logger.severe(
-          """Error formatting the generated source code.
-This may indicate an issue in the generated code or in the formatter.
-Please check the generated code and file an issue on source_gen
-if approppriate.""",
+          ''
```
natebosch: [nit] remove

kevmoo: Blank string huh?

matanlurey: I actually put this so the entire rest of the message is not formatted strangely - it's sort of a cludge with dartfmt. I can remove if you like, but then the error message doesn't stick into one paragraph.

lib/src/builder.dart
```diff
@@ -78,10 +78,13 @@ class GeneratorBuilder extends Builder {
...
+          'Error formatting generated source code for ${library.identifier} '
+          'which was output to ${_generatedFile(buildStep.inputId).path}.\n'
+          'This may indicate an issue in the generated code or in the formatter.'
+          '\n'
```
kevmoo: Consider using `'''...'''` for multi-line things.

lib/src/builder.dart
matanlurey: Done

lib/src/builder.dart
matanlurey: Done

natebosch: We've been moving to this format in a lot of places. `'''` works best for things like top-level variables where the string doesn't have any long lines

- let's us break up the String arbitrarily and and separate from how it's broken in the output to fit 80 chars
- Doesn't introduce a weird jump to the left of the screen from an indented block when the output *shouldn't* be indented

matanlurey:
@kevmoo I sent a cooresponding CL to patch this internally.

When you eventually sync, this will safely override it. If that's OK, I'll let you release normally.

kevmoo:
@matanlurey – hrm...not a good practice to start IMHO – very easy to publish this and do a normal sync...and avoid confusing folks maintaining 3rd-party

matanlurey:
Sure, but we're time sensitive right now. Given that @natebosch supports the package internally too I don't think it's a huge issue. Normally I would just wait for the sync.

lib/src/builder.dart
```diff
@@ -78,10 +78,10 @@ class GeneratorBuilder extends Builder {
       genPartContent = formatter.format(genPartContent);
     } catch (e, stack) {
       buildStep.logger.severe(
-          """Error formatting the generated source code.
-This may indicate an issue in the generated code or in the formatter.
-Please check the generated code and file an issue on source_gen
-if approppriate.""",
+          '''Error formatting generated source code for ${library.identifier}
+which was output to ${_generatedFile(buildStep.inputId).path}.\n
```
natebosch: We ended up with a bad mix here - we don't need both `'''` and `\n`...

PR #133, opened 2017-02-13
Reflow and format help message

natebosch:
Use concatenated `''` rather than `'''` so that we can break lines
separately in the source from where we want them broken in the output


natebosch:
Once I merge this I can pull it internally as well.

PR #138, opened 2017-03-31
Refer to library by URL.

davidmorgan:
This will let people get rid of 'library' statements.

Really looking forward to this change -- any plans for when we can make it? Do we need to wait for 1.23 to be in the stable channel?

I checked built_value's end to end tests with 1.23.0-dev.11.0 and it works fine. The analyzed currently complains, but we don't necessarily need to wait for that -- we can silence it with "// ignore: non_identifier_library_name".

Thanks!

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -4,7 +4,7 @@ author: Dart Team <misc@dartlang.org>
 description: Automatic sourcecode generation for Dart
 homepage: https://github.com/dart-lang/source_gen
 environment:
-  sdk: '>=1.12.0 <2.0.0'
+  sdk: '>=1.23.0-dev.11.0 <2.0.0'
 dependencies:
   analyzer: ^0.29.2
   build: ^0.7.1
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -8,7 +8,7 @@ environment:
 dependencies:
   analyzer: ^0.29.2
   build: ^0.7.1
-  dart_style: '>=0.1.7 <2.0.0'
+  dart_style: ^1.0.0
   path: ^1.3.2
 dev_dependencies:
   build_runner: ^0.3.0
--- a/test/builder_test.dart
+++ b/test/builder_test.dart
@@ -153,13 +153,15 @@ class MyGoodError { }
 ''';
 
 const _testLibPartContent = r'''
-part of test_lib;
+// ignore: non_identifier_library_name
+part of 'test_lib.dart';
 final int bar = 42;
 class Customer { }
 ''';
 
 const _testGenPartContent = r'''// GENERATED CODE - DO NOT MODIFY BY HAND
 
+// ignore: non_identifier_library_name
 part of 'test_lib.dart';
 
 
@@ -181,6 +183,7 @@ part of 'test_lib.dart';
 const _testGenPartContentForLibrary =
     r'''// GENERATED CODE - DO NOT MODIFY BY HAND
 
+// ignore: non_identifier_library_name
 part of 'test_lib.dart';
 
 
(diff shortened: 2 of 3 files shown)


pubspec.yaml
```diff
@@ -4,7 +4,7 @@ author: Dart Team <misc@dartlang.org>
 description: Automatic sourcecode generation for Dart
 homepage: https://github.com/dart-lang/source_gen
 environment:
-  sdk: '>=1.12.0 <2.0.0'
+  sdk: '>=1.23.0-dev.11.0 <2.0.0'
```
kevmoo: Is this about the SDK min value ... or pkg/analyzer min version?

jakemac53: Probably need both? SDK for supporting the output in tools/runtimes,  analyzer for supporting parsing this file in later codegen steps etc?

kevmoo: Strictly speaking, this package doesn't *need* to align w/ the SDK version. Just the analyzer version.

One could generate code with SDK X that doesn't run in SDK X.

...right?

davidmorgan: Yes, we could generate code that doesn't run. I guess if we do that we would instead need to bump the source_gen major version? Maybe we should do that anyway?

--> no new SDK requirement
--> version bump to 0.6

What do you think?

davidmorgan: Friendly ping. Any ideas how we should ship this? Thanks.

jakemac53: I think we should bump the min sdk version but not do a major version bump in this package (unless there is a breaking change I am missing).

There isn't much of a point in generating code that can't run in your current sdk imo, it would just lead to confusion.

test/builder_test.dart
```diff
@@ -160,7 +160,8 @@ class Customer { }
 
 const _testGenPartContent = r'''// GENERATED CODE - DO NOT MODIFY BY HAND
 
-part of test_lib;
+part of 'test_lib.dart';
+
```
kevmoo: where is this extra new-line coming from?

test/builder_test.dart
```diff
@@ -160,7 +160,8 @@ class Customer { }
 
 const _testGenPartContent = r'''// GENERATED CODE - DO NOT MODIFY BY HAND
 
-part of test_lib;
+part of 'test_lib.dart';
```
kevmoo: Also: We could remove the library name from `test_lib.dart`, right?

test/builder_test.dart
davidmorgan: :/ don't know ... it may be some weirdness with dartfmt.

test/builder_test.dart
davidmorgan: Yes. If we want to push this before the analyzer is fixed, we'll also want to generate the appropriate '// ignore: '. WDYT?

jakemac53: Lets just generate it with the `// ignore: ` for now.

pubspec.yaml
davidmorgan: Ack, kept it with just the SDK version bump.

test/builder_test.dart
davidmorgan: Added the "ignore".

davidmorgan:
There seems to be an issue with dartfmt; it appears to be refusing to format the output altogether.

When I run dartfmt locally, it actually deletes the URL from 'part of'. So I think this CL will have to wait for a dartfmt fix. I'll see if there's an issue open.

davidmorgan:
https://github.com/dart-lang/dart_style/issues/615

matanlurey:
@davidmorgan Ping. This was fixed as of May 9th 👍

matanlurey:
@davidmorgan I'm proposing to supersede this with https://github.com/dart-lang/source_gen/pull/179.

PR #141, opened 2017-04-14
Support package:build 0.8.x

natebosch:
There are no breaking changes as far as source_gen is concerned

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -11,7 +11,7 @@ dependencies:
   dart_style: '>=0.1.7 <2.0.0'
   path: ^1.3.2
 dev_dependencies:
-  build_runner: ^0.3.0
+  build_runner: ^0.3.1
   build_test: ^0.5.1
   cli_util: ^0.0.1
   collection: ^1.1.2
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -11,7 +11,7 @@ dependencies:
   dart_style: '>=0.1.7 <2.0.0'
   path: ^1.3.2
 dev_dependencies:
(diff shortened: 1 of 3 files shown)


kevmoo:
Are you sure?

https://travis-ci.org/dart-lang/source_gen/jobs/222178146#L178

natebosch:
Limited by dev_dependencies today. We can't bump `build_runner` yet.

If you want we can hold off on this change until then - but this should unblock consumers of this package from using the new build since they'll ignore dev_dependencies.

natebosch:
Holding off for now until we bump the dev_dependencies

kevmoo:
Don't make travis sad!

pubspec.yaml
```diff
@@ -1,18 +1,18 @@
 name: source_gen
-version: 0.5.4+3
+version: 0.5.4+4
 author: Dart Team <misc@dartlang.org>
 description: Automatic sourcecode generation for Dart
 homepage: https://github.com/dart-lang/source_gen
 environment:
   sdk: '>=1.12.0 <2.0.0'
```
kevmoo: Do we want to bump this, too?

natebosch: I will end up doing that with https://github.com/dart-lang/source_gen/pull/142

Would prefer for it to happen there, because that is the commit which makes the code need 1.21 rather than just deps (which pub will be checking for us)

natebosch:
I have finally appeased Travis

pubspec.yaml
kevmoo: kk

PR #142, opened 2017-04-17
Less verbose error for analysis error

natebosch:
The AnalysisException thrown during initial iterating over elements can
be very noisy. Stack traces are included in the `toString()` and it's
detail that most users won't care about.

- Add a safeIterate utility method which allows catching the exceptions
  specifically thrown by the elements iterable and not by the Builder
- Simplify to a message about failing to resolve the input. Include the
  extra details in a log.fine

### DIFF
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -100,8 +100,8 @@ class GeneratorBuilder extends Builder {
 Stream<GeneratedOutput> _generate(LibraryElement unit,
     List<Generator> generators, BuildStep buildStep) async* {
   var elements = safeIterate(getElementsFromLibraryElement(unit), (e, [_]) {
-    log.severe('Failed to resolve ${buildStep.inputId}.');
     log.fine('Resolve error details:\n$e');
+    log.severe('Failed to resolve ${buildStep.inputId}.');
   });
   for (var element in elements) {
     yield* _processUnitMember(element, generators, buildStep);
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -4,6 +4,7 @@
 import 'dart:async';
 
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/exception/exception.dart';
 import 'package:build/build.dart';
 import 'package:dart_style/src/dart_formatter.dart';
 
@@ -98,7 +99,11 @@ class GeneratorBuilder extends Builder {
 
 Stream<GeneratedOutput> _generate(LibraryElement unit,
     List<Generator> generators, BuildStep buildStep) async* {
-  for (var element in getElementsFromLibraryElement(unit)) {
+  var elements = safeIterate(getElementsFromLibraryElement(unit), (e, [_]) {
+    log.severe('Failed to resolve ${buildStep.inputId}.');
+    log.fine('Resolve error details:\n$e');
+  });
+  for (var element in elements) {
     yield* _processUnitMember(element, generators, buildStep);
   }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ dart:
   - stable
   - 1.22.1
   - 1.21.1
-  - 1.20.1
-  - 1.19.1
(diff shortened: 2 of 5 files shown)


lib/src/builder.dart
```diff
@@ -98,7 +99,11 @@ class GeneratorBuilder extends Builder {
 
 Stream<GeneratedOutput> _generate(LibraryElement unit,
     List<Generator> generators, BuildStep buildStep) async* {
-  for (var element in getElementsFromLibraryElement(unit)) {
```
kevmoo: Before, this would just blow up.

Now it'll succeed, but log errors, right? Or are we treating `severe` logs as fatal?

Just trying to understand...

natebosch: `severe` logs are fatal in bazel and build_barback.
They are tolerated in build_runner but we want to change that: https://github.com/dart-lang/build/issues/215

lib/src/builder.dart
```diff
@@ -98,7 +99,11 @@ class GeneratorBuilder extends Builder {
...
+  var elements = safeIterate(getElementsFromLibraryElement(unit), (e, [_]) {
+    log.severe('Failed to resolve ${buildStep.inputId}.');
+    log.fine('Resolve error details:\n$e');
```
jakemac53: isn't this dead code since severe is fatal?

lib/src/builder.dart
```diff
@@ -98,7 +99,11 @@ class GeneratorBuilder extends Builder {
 
 Stream<GeneratedOutput> _generate(LibraryElement unit,
     List<Generator> generators, BuildStep buildStep) async* {
-  for (var element in getElementsFromLibraryElement(unit)) {
+  var elements = safeIterate(getElementsFromLibraryElement(unit), (e, [_]) {
+    log.severe('Failed to resolve ${buildStep.inputId}.');
```
jakemac53: Should we add more info here? What error is this actually indicating?

lib/src/builder.dart
```diff
@@ -98,7 +99,11 @@ class GeneratorBuilder extends Builder {
 
 Stream<GeneratedOutput> _generate(LibraryElement unit,
     List<Generator> generators, BuildStep buildStep) async* {
-  for (var element in getElementsFromLibraryElement(unit)) {
+  var elements = safeIterate(getElementsFromLibraryElement(unit), (e, [_]) {
```
jakemac53: It seems like in general we could just handle this inside of `getElementsFromLibraryElement` or `_getElements` (whichever is throwing).

natebosch: We could. It would introduce a `package:build` import into `utils.darts` so we could do the logging, and we'd probably end up needing to pass through the `buildStep` or `inputId` which IMO degrades from the nice separation we have now.

My preference would be this layout, WDYT?

lib/src/builder.dart
natebosch: It could mean a lot of things which is why I kept it so general. Up until now we encounter this mostly because of our bazel codegen not correctly handling things - quirks in build targets etc.

I could add something like "File an issue" but the details of how and where to file will likely change by situation...

lib/src/builder.dart
natebosch: It's no longer synchronously fatal.
This log does show up in bazel with --define=DART_CODEGEN_LOG_LEVE=fine

Would you prefer if I move it earlier than the log.severe for safety?

jakemac53: Ya lets just move it up one line even though that is a bit weird.

lib/src/builder.dart
jakemac53: kk, sgtm

lib/src/builder.dart
jakemac53: I am OK with it as is, that would be somewhat annoying for sure.

lib/src/builder.dart
natebosch: Done

PR #145, opened 2017-04-25
Upgrade to build 0.9.0

natebosch:
- Replace declareOutputs with buildExtensions
- Upgrade build related dependencies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ dart:
   - dev
   - stable
   - 1.22.1
-  - 1.21.1
 cache:
   directories:
     - $HOME/.pub-cache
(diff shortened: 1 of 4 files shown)


kevmoo:
FYI, @natebosch – might could have updated the min SDK constraint to `1.21.1` as well...

natebosch:
Min SDK constraint was already 1.21.1

kevmoo:
Er 1.22.1

From mobile. Sorry for brevity and/or spelling.

natebosch:
Since the SDK constraint is tightened in package:build rather than by usage here I don't think it makes sense to update this constraint - we let pub figure that out for us. IMO it doesn't make sense to transitively search a package's dependencies to find the SDK constraint but to specify it based on language usage within this package.

kevmoo:
>  IMO it doesn't make sense to transitively search a package's dependencies to find the SDK constraint but to specify it based on language usage within this package.

For me it's about support. If we can't test version X, we shouldn't claim to support it.

PR #147, opened 2017-06-01
Add option to do custom code formatting

jakemac53:


jakemac53:
Updating this to instead take `String formatCode(String generatedCode)` argument so you can do whatever you want, including just returning the original string.

natebosch:
Who is asking for this? Is the motivation a performance improvement?

jakemac53:
The motivation is performance primarily yes. The angular compiler creates some code which dartfmt takes a long time to figure out (hundreds of milliseconds on some real examples).

In that case the solution was to increase the line length, but we opted for a more general solution which allows any custom formatting of code.

PR #149, opened 2017-06-13
Lints

kevmoo:


.travis.yml
```diff
@@ -3,21 +3,24 @@ sudo: false
 dart:
   - dev
   - stable
+  - 1.23.0
```
jakemac53: Are these really necessary? Seems like just dev/stable + lowest bound for environment in the pubspec would be sufficient.

kevmoo: Eh, not a bad idea. Just testing the lower bound seams reasonable.

kevmoo:
PTAL, @jakemac53 – updated SDK constraint

PR #150, opened 2017-06-13
Serialization helpers

kevmoo:


kevmoo:
@jakemac53 builds off https://github.com/dart-lang/source_gen/pull/149

lib/generators/json_serializable_generator.dart
```diff
@@ -2,23 +2,25 @@
...
 
+export 'package:source_gen/src/json_serializable/type_helper.dart';
+
 // TODO: toJson option to omit null/empty values
 class JsonSerializableGenerator
     extends GeneratorForAnnotation<JsonSerializable> {
-  const JsonSerializableGenerator();
+  final List<TypeHelper> typeHelpers;
+
+  const JsonSerializableGenerator(
```
jakemac53: I would also add a top level `defaultHelpers` list or a `withDefaultHelpers` constructor so people don't have to hardcode the list of default ones if they supply custom ones. That way if we add more default helpers they can opt into getting those by default.

lib/generators/json_serializable_generator.dart
```diff
@@ -48,7 +50,8 @@ class JsonSerializableGenerator
     var buffer = new StringBuffer();
 
     if (annotation.createFactory) {
-      var toSkip = _writeFactory(buffer, classElement, fields, prefix);
+      var toSkip =
+          writeFactory(buffer, classElement, fields, prefix, typeHelpers);
```
jakemac53: nit: Imo `writeFactory` should actually be a private method in this class (and all the other ones like it). They are only used by this class afaict and you could remove the `typeHelpers` arg that way.

lib/src/json_serializable/type_helper.dart
```diff
@@ -0,0 +1,83 @@
...
+import 'package:analyzer/dart/element/type.dart';
+
+abstract class TypeHelper {
+  const TypeHelper();
+
+  bool matchesTypeForSerialize(DartType type);
+
+  bool canDeserialize(DartType type);
+
+  String toSerializedValue(DartType searchType, String name);
```
jakemac53: Lets add comments to these methods, at least toSerializedValue and toDeserializedValue. I was confused why they took Strings and returned Strings until I got to the implementations and realized they were returning actual code to be written out as a String.

lib/src/json_serializable/type_helper.dart
```diff
@@ -0,0 +1,83 @@
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+
+abstract class TypeHelper {
+  const TypeHelper();
+
+  bool matchesTypeForSerialize(DartType type);
```
jakemac53: I would unify these names, either change this to `canSerialize` or change `canDeserialize` to `matchesTypeForDeserialize`.

lib/src/json_serializable/type_helper.dart
```diff
@@ -0,0 +1,83 @@
...
+
+class JsonHelper extends TypeHelper {
+  const JsonHelper();
+
+  @override
+  bool matchesTypeForSerialize(DartType type) {
+    //TODO(kevmoo): This should be checking for toJson method, but
+    //              toJson might be gone during generation, so check for
+    //              the annotation, too!
+    if (type is! InterfaceType) return false;
```
jakemac53: Since this is identical to `canDeserialize` right now I would just call it - that also makes it more clear what this workaround is doing.

lib/src/json_serializable/type_helper.dart
```diff
@@ -0,0 +1,83 @@
...
+        // TODO: validate that there are the right number and type of arguments
+        return true;
+      }
+    }
+
+    return false;
+  }
+
+  @override
+  String toSerializedValue(DartType searchType, String name) => name;
```
jakemac53: So is this just relying on the fact that default json encoding will call toJson? That seems like it is ok but I would like to confirm my understanding. Maybe it is worth adding a comment as well?

lib/generators/json_serializable_generator.dart
kevmoo: ack

lib/src/json_serializable/type_helper.dart
kevmoo: Yup.

lib/src/json_serializable/type_helper.dart
kevmoo: Ah! good idea!

lib/src/json_serializable/type_helper.dart
kevmoo: done

lib/generators/json_serializable_generator.dart
kevmoo: done

kevmoo:
sorry for the force squash – had a rebase mess to deal with.

PTAL

lib/generators/json_serializable_generator.dart
```diff
@@ -2,23 +2,37 @@
...
 // TODO: toJson option to omit null/empty values
 class JsonSerializableGenerator
     extends GeneratorForAnnotation<JsonSerializable> {
-  const JsonSerializableGenerator();
+  static const _defaultHelpers = const [
+    const JsonHelper(),
+    const DateTimeHelper()
+  ];
+
+  final List<TypeHelper> typeHelpers;
```
jakemac53: Type this as `Iterable` IMO since its always const or unmodifiable.

jakemac53: And/or, make it private?

lib/src/json_serializable/type_helper.dart
```diff
@@ -0,0 +1,112 @@
...
+  /// String serialize(DartType targetType, String expression) =>
+  ///   "$expression.id";
+  /// ```.
+  String serialize(DartType targetType, String expression);
+
+  /// Returns Dart code that deserializes an [expression] representing a JSON
+  /// literal to into [targetType].
+  ///
+  /// Let's say you want to deserialize a class `Foo` by taking an `int` stored
+  /// in a JSON literal and calling they `Foo.fromInt` constructor.
```
jakemac53: they -> the

lib/generators/json_serializable_generator.dart
kevmoo: done

lib/src/json_serializable/type_helper.dart
kevmoo: done

natebosch:
Why are investing in the JSON serializers? I've been hoping to find the time to kill them or move them to a new package.

Who is using these?

kevmoo:
> Who is using these?

I am 😄 

They are a great integration test, IMHO.

I am a fan of moving them into their own package.

Issue #136, opened 2017-02-16
Consider helpers for static type reflection


matanlurey:
Within Angular today I'm working on a class called `StaticTypes`:

```dart
/// An abstraction around comparing types statically (at compile-time).
abstract class StaticTypes {
  /// Creates a [MetadataTypes] that uses runtime reflection.
  const factory StaticTypes.withMirrors({
    bool checkSubTypes,
  }) = _MirrorsStaticTypes;

  /// Creates a [MetadataTypes] by retrieving exact types from [resolver].
  ///
  /// With _analysis summaries_ this should be a relatively cheap operation but
  /// the resulting [MetadataTypes] class should still be cached whenever
  /// possible.
  ///
  /// Throws an [ArgumentError] if [resolver] was not able to lookup type
  /// information for Angular's metadata classes - this is usually a sign that
  /// dependencies or imports are missing.
  factory StaticTypes.fromResolver(
    Resolver resolver, {
    bool checkSubTypes,
  }) = ...
}
```

This could be useful to all users that are writing source-code generation/macros.

Would be happy to port/commonify this once complete. It could replace the hackier 'matchAnnotation' that is used within the `src/` folder today.

matanlurey:
Ping @kevmoo @natebosch @jakemac53 - is this desirable?

jakemac53:
Could you give a bit more information about this class, for instance what
instance methods do your forsee? Is it used to check if  an analyzer
`DartObject` is of a certain real `Type`?

matanlurey:
Yep! Incidentally I'm in SEA today so I can show it off

PR #153, opened 2017-06-18
Add TypeChecker class.

matanlurey:
This will be nice to have versus stuffed into the AngularDart compiler, and it doesn't feel "right" going into a library like `build`, which I think benefits from having a very small interface and low-threshold for changes.

Advantages of this approach
* Adds an "approved" (not `/lib/src/annotations.dart') way of performing checks.
* Has two backing implementations not based on `dart:mirrors` (for long-term safety).
* Easy re-usable setup for complex compilers like `inject` or `angular`.

See `type_checker_test.dart` for some functional tests.

Closes https://github.com/dart-lang/source_gen/issues/136.

### DIFF
--- a/lib/src/type_checker.dart
+++ b/lib/src/type_checker.dart
@@ -113,6 +113,9 @@ class _MirrorTypeChecker extends TypeChecker {
 // Checks a runtime type against an Uri and Symbol.
 class _UriTypeChecker extends TypeChecker {
   static String _urlOf(Element element) {
+    if (element.kind == ElementKind.DYNAMIC) {
+      return 'dart:core#dynamic';
+    }
     var sourceUri = element.source.uri;
     switch (sourceUri.scheme) {
       case 'dart':
--- a/lib/src/type_checker.dart
+++ b/lib/src/type_checker.dart
@@ -34,6 +34,9 @@ abstract class TypeChecker {
   ///
   /// Otherwise returns `null`.
   DartObject firstAnnotationOf(Element element) {
+    if (element.metadata.isEmpty) {
+      return null;
+    }
     final results = annotationsOf(element);
     return results.isEmpty ? null : results.first;
   }
@@ -57,29 +60,50 @@ abstract class TypeChecker {
   bool isSuperTypeOf(DartType staticType) => isSuperOf(staticType.element);
 }
 
-String _urlOf(Element element) {
-  if (element.kind == ElementKind.DYNAMIC) {
-    return 'dart:core#dynamic';
-  }
-  var sourceUri = element.source.uri;
-  switch (sourceUri.scheme) {
+Uri _normalizeUrl(Uri url) {
+  switch (url.scheme) {
     case 'dart':
-      // Some internal dart: URLs are something like dart:core/map.dart.
-      //
-      // This isn't a user-knowable path, so we strip out extra path segments
-      // and only expose dart:core.
-      if (sourceUri.pathSegments.isNotEmpty) {
-        final path = sourceUri.pathSegments.first;
-        sourceUri = sourceUri.replace(pathSegments: [path]);
-      }
-      break;
+      return _normalizeDartUrl(url);
     case 'package':
-      break;
+      return _packageToAssetUrl(url);
     default:
-      throw new StateError('Cannot resolve "$sourceUri".');
+      return url;
   }
-  return sourceUri.replace(fragment: element.name).toString();
 }
+
+/// Make `dart:`-type URLs look like a user-knowable path.
+///
+/// Some internal dart: URLs are something like `dart:core/map.dart`.
+///
+/// This isn't a user-knowable path, so we strip out extra path segments
+/// and only expose `dart:core`.
+Uri _normalizeDartUrl(Uri url) => url.pathSegments.isNotEmpty
+    ? url.replace(pathSegments: url.pathSegments.take(1))
+    : url;
+
+/// Returns a `package:` URL into a `asset:` URL.
+///
+/// This makes internal comparison logic much easier, but still allows users
+/// to define assets in terms of `package:`, which is something that makes more
+/// sense to most.
+///
+/// For example this transforms `package:source_gen/source_gen.dart` into:
+/// `asset:source_gen/lib/source_gen.dart`.
+Uri _packageToAssetUrl(Uri url) => url.scheme == 'package'
+    ? url.replace(
+        scheme: 'asset',
+        pathSegments: <String>[]
+          ..add(url.pathSegments.first)
+          ..add('lib')
+          ..addAll(url.pathSegments.skip(1)))
+    : url;
+
+/// Returns
+String _urlOfElement(Element element) => element.kind == ElementKind.DYNAMIC
+    ? 'dart:core#dynmaic'
+    : _normalizeUrl(element.source.uri)
+        .replace(fragment: element.name)
+        .toString();
 
 // Checks a static type against another static type;
 class _LibraryTypeChecker extends TypeChecker {
@@ -92,25 +116,14 @@ class _LibraryTypeChecker extends TypeChecker {
       element is ClassElement && element == _type.element;
 
   @override
-  String toString() => '${_urlOf(_type.element)}';
+  String toString() => '${_urlOfElement(_type.element)}';
 }
 
 // Checks a runtime type against a static type.
 class _MirrorTypeChecker extends TypeChecker {
-  static Uri _uriOf(ClassMirror mirror) {
-    final sourceUri = (mirror.owner as LibraryMirror).uri;
-    switch (sourceUri.scheme) {
-      case 'dart':
-      case 'package':
-        return sourceUri.replace(
-          fragment: MirrorSystem.getName(mirror.simpleName),
-        );
-      default:
-        throw new StateError(
-            'Cannot resolve "$sourceUri". You must import ${mirror.simpleName} '
-            'class using a package: or dart: url.');
-    }
-  }
+  static Uri _uriOf(ClassMirror mirror) =>
+      _normalizeUrl((mirror.owner as LibraryMirror).uri).replace(
+          fragment: MirrorSystem.getName(mirror.simpleName));
 
   // Precomputed type checker for types that already have been used.
   static final _cache = new Expando<TypeChecker>();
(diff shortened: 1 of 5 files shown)


thosakwe:
I just saw this PR. This might also close #155 (which I just opened). I'll have to check...

kevmoo:
@matanlurey I like...but!

I noticed when trying to convert the JsonSerialiableGenerator to it it blows up w/ the `dynamic` element – which (for some reason) has no source.

If you could – check it out. Or I'll do a follow-up. It's certainly better than the hacks I have in place to detect DateTime, Iterable, and List

matanlurey:
@kevmoo Let me try converting your generator as a test case then.

kevmoo:
Look at this crazy
https://github.com/dart-lang/source_gen/blob/master/lib/generators/json_serializable_generator.dart#L318

matanlurey:
@kevmoo Just updated, and added `annotationOf` which seems to work fine:

```dart
String _fieldToJsonMapKey(String fieldName, FieldElement field) {
  const $JsonKey = const TypeChecker.fromRuntime(JsonKey);
  var jsonKey = $JsonKey.annotationOf(field);
  if (jsonKey != null) {
    var jsonName = jsonKey.getField('jsonName').toStringValue();
    return jsonName;
  }
  return fieldName;
}
```

lib/src/type_checker.dart
```diff
@@ -0,0 +1,152 @@
...
+  /// const linkedHashMap = const TypeChecker.fromUrl(
+  ///   'dart:collection#LinkedHashMap',
+  /// );
+  /// ```
+  const factory TypeChecker.fromUrl(dynamic url) = _UriTypeChecker;
+
+  /// Returns the first constant annotating [element] that is this type.
+  ///
+  /// Otherwise returns `null`.
+  DartObject annotationOf(Element element) {
```
jakemac53: Should this be `firstAnnotationOf`? And then possibly add `annotationsOf`?

lib/src/type_checker.dart
```diff
@@ -0,0 +1,152 @@
...
+      final constant = metadata.computeConstantValue();
+      if (isExactlyType(constant.type)) {
+        return constant;
+      }
+    }
+    return null;
+  }
+
+  /// Returns `true` if representing the exact same class as [element].
+  bool isExactly(Element element);
```
jakemac53: Should this be `ClassElement`?

lib/src/type_checker.dart
```diff
@@ -0,0 +1,152 @@
...
+  /// Returns `true` if representing the exact same type as [staticType].
+  bool isExactlyType(DartType staticType) => isExactly(staticType.element);
+
+  /// Returns `true` if representing a super class of [element].
+  bool isSuperOf(Element element) =>
+      isExactly(element) ||
+      element is ClassElement && element.allSupertypes.any(isExactlyType);
+
+  /// Returns `true` if representing a super type of [staticType].
+  bool isSuperType(DartType staticType) => isSuperOf(staticType.element);
```
jakemac53: nit: `isSuperTypeOf` for consistency

lib/src/type_checker.dart
```diff
@@ -0,0 +1,152 @@
...
+
+import 'dart:mirrors';
+
+import 'package:analyzer/dart/constant/value.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+
+/// An abstraction around doing static type checking at compile/build time.
+abstract class TypeChecker {
+  const TypeChecker._();
```
jakemac53: Do you need this constructor? You should be able to delete it and remove the super constructor calls.

lib/src/type_checker.dart
```diff
@@ -0,0 +1,152 @@
...
+
+  @override
+  bool isExactly(Element element) =>
+      element is ClassElement && element.type == _type;
+
+  @override
+  bool isSuperOf(Element element) =>
+      isExactly(element) ||
+      element is ClassElement &&
+          element.allSupertypes.any((t) => t.element == _type.element);
```
jakemac53: Why is the logic different for checking super types? (here it checks for ClassElement equality where `isExactly` checks for DartType equality)

lib/src/type_checker.dart
```diff
@@ -0,0 +1,152 @@
...
+
+  @override
+  String toString() => '${_UriTypeChecker._urlOf(_type.element)}';
```
jakemac53: nit: consider moving `urlOf` to a top level method if its going to be used by more than one implementation

lib/src/type_checker.dart
```diff
@@ -0,0 +1,152 @@
...
+  static Uri _uriOf(ClassMirror mirror) {
+    final sourceUri = (mirror.owner as LibraryMirror).uri;
+    switch (sourceUri.scheme) {
+      case 'dart':
+      case 'package':
+        return sourceUri.replace(
+          fragment: MirrorSystem.getName(mirror.simpleName),
+        );
+      default:
+        throw new StateError('Cannot resolve "$sourceUri".');
```
jakemac53: maybe add an additional note `You must import the ${mirror.simpleName} class using a package: or dart: uri.`.

lib/src/type_checker.dart
```diff
@@ -0,0 +1,152 @@
...
+// Checks a runtime type against an Uri and Symbol.
+class _UriTypeChecker extends TypeChecker {
+  static String _urlOf(Element element) {
+    if (element.kind == ElementKind.DYNAMIC) {
+      return 'dart:core#dynamic';
+    }
+    var sourceUri = element.source.uri;
+    switch (sourceUri.scheme) {
+      case 'dart':
+        if (sourceUri.pathSegments.isNotEmpty) {
```
jakemac53: Is it an error if this is empty? Also when is it more than one segment?

test/type_checker_test.dart
```diff
@@ -0,0 +1,98 @@
...
+  DartType staticHashMap;
+  TypeChecker staticMapChecker;
+  TypeChecker staticHashMapChecker;
+
+  setUpAll(() async {
+    final resolver = await resolveSource('');
+
+    final core = resolver.getLibraryByName('dart.core');
+    staticMap = core.getType('Map').type;
+    staticMapChecker = new TypeChecker.fromStatic(staticMap);
```
jakemac53: Might as well just create this inline in the test that uses them below like the others instead of creating them for all tests (same for `staticHashMapChecker`).

test/type_checker_test.dart
```diff
@@ -0,0 +1,98 @@
...
+      test('should equal dart:core#Map', () {
+        expect(
+          checkMap().isExactlyType(staticMap),
+          isTrue,
+          reason: '$checkMap != $staticMap',
+        );
+      });
+
+      test('should not be a super type of dart:core#Map', () {
+        expect(checkMap().isSuperType(staticMap), isTrue);
```
jakemac53: Shouldn't this be `isFalse`?

test/type_checker_test.dart
```diff
@@ -0,0 +1,98 @@
...
+  void commonTests({
+    @required TypeChecker checkMap(),
+    @required TypeChecker checkHashMap(),
+  }) {
+    group('(Map)', () {
+      test('should equal dart:core#Map', () {
+        expect(
+          checkMap().isExactlyType(staticMap),
+          isTrue,
+          reason: '$checkMap != $staticMap',
```
jakemac53: nit: no trailing comma ;) (also below).

kevmoo:
See broken build – `Cannot resolve "asset:source_gen/lib/generators/json_serializable.dart".`

Ugh...

lib/src/type_checker.dart
matanlurey: Done.

lib/src/type_checker.dart
matanlurey: Fixed, thanks!

lib/src/type_checker.dart
matanlurey: Done.

lib/src/type_checker.dart
matanlurey: Done.

lib/src/type_checker.dart
matanlurey: Added a comment why.

test/type_checker_test.dart
matanlurey: The only reason I did this is to avoid creating a `Resolver` multiple times per test.

test/type_checker_test.dart
matanlurey: Done.

test/type_checker_test.dart
matanlurey: Done and fixed.

lib/src/type_checker.dart
jakemac53: Ah ok just for the const, makes sense. We can always make this public later if somebody actually has a valid reason to extend it and files an issue.

test/type_checker_test.dart
jakemac53: oh right, `setUpAll` just runs once doesn't it, sgtm

lib/src/type_checker.dart
jakemac53: ok, sgtm

lib/src/type_checker.dart
jakemac53: where? (maybe forgot to push a commit?)

kevmoo:
@matanlurey thoughts on the broken travis bits?

matanlurey:
@kevmoo Fixing now, I had some mismatch b/w `asset:` and `package:` URLs.

New functionality normalizes the two (internally). Will upload as soon as all tests pass.

kevmoo:
@matanlurey for this PR or later – would be great to have a `isExactlyOrIsSuperTypeOf` – or similar.

See https://github.com/dart-lang/source_gen/pull/164/commits/0046bdd00c2c1b5a57f0d730535e72cdf71ea5c0#diff-c4274d5939b2f855cfba15318f2d42ebR250

Having to do `isExact || isSuper` several places

matanlurey:
@kevmoo Send a PR after I merge? I finally got this one green :P

kevmoo:
> Send a PR after I merge? I finally got this one green :P

Deal. 👍

Issue #135, opened 2017-02-16
Clarify role of this package v package:build


matanlurey:
It's not clear to users on gitter commonly.

Perhaps this should be more specific too, like `package:macro` or something like that? Dunno.

kevmoo:
This is an opinionated way to generate code via annotations.

...but it would be good to clarify.

PR #156, opened 2017-06-18
Incremental enhancements to README and dartdoc

matanlurey:
Specifically, clarifying the _role_ of this package (since it forked into `build`), what this package _does_ and (does not do), and added API documentation to the most common classes that were sorely lacking.

I also added a gitter chatroom, since there seems to be enough users trying this out to warrant. We can probably re-use this chatroom for the various build packages.

Closes https://github.com/dart-lang/source_gen/issues/135.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,56 @@
-[![Build Status](https://travis-ci.org/dart-lang/source_gen.svg?branch=master)](https://travis-ci.org/dart-lang/source_gen)
+# source_gen
 
-`source_gen` provides:
+<p align="center">
+  <a href="https://travis-ci.org/dart-lang/source_gen">
+    <img src="https://travis-ci.org/dart-lang/source_gen.svg?branch=master" alt="Build Status" />
+  </a>
+  <a href="https://pub.dartlang.org/packages/source_gen">
+    <img src="https://img.shields.io/pub/v/twitch.svg" alt="Pub Package Version" />
+  </a>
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,13 +1,57 @@
-[![Build Status](https://travis-ci.org/dart-lang/source_gen.svg?branch=master)](https://travis-ci.org/dart-lang/source_gen)
+# source_gen
 
-`source_gen` provides:
+<p align="center">
```
jakemac53: Any idea what this looks like on the pub site or other areas? Not sure if its worth this potential headache just to get them centered :).

README.md
```diff
@@ -1,13 +1,57 @@
...
+If you're only using `source_gen` in your own project to generate code, and
+then you publish that code (generated code included), then you can simply add
+as a `dev_dependency`:
+
+```yaml
+dev_dependencies:
+  source_gen:
+```
+
+[pub_deps]: https://www.dartlang.org/tools/pub/dependencies
```
jakemac53: nit: move this below with the other ones?

README.md
matanlurey: I double checked this renders fine.

README.md
matanlurey: I'm going to do a few more additions if its OK to fix upstream.

natebosch:
:+1: 

Thanks!

Issue #155, opened 2017-06-18
matchAnnotation and instantiateAnnotation break on extended classes


thosakwe:
Hi, I'm experiencing an error that is blocking my progress using `package:source_gen`. I am more than happy to submit a PR to fix it, but I'd need clearance first.

Anyways, I'm building a PostgreSQL ORM generator, driven by a few annotations. I chose to model relations the following way:

```dart
class Relationship {
  final String localKey;
  final String foreignKey;
  final String foreignTable;
  final bool cascadeOnDelete;

  const Relationship._(
      {this.localKey,
      this.foreignKey,
      this.foreignTable,
      this.cascadeOnDelete});
}

class HasMany extends Relationship {
  const HasMany(
      {String localKey,
      String foreignKey,
      String foreignTable,
      bool cascadeOnDelete: false})
      : super._(
            localKey: localKey,
            foreignKey: foreignKey,
            foreignTable: foreignTable,
            cascadeOnDelete: cascadeOnDelete == true);
}

const HasMany hasMany = const HasMany();  
// HasOne and BelongsTo (omitted here) are virtually identical to HasMany
```

I've written a helper `findAnnotation<T>` function that calls `matchAnnotation` and `instantiateAnnotation` in one go.

```dart
import 'package:analyzer/dart/element/element.dart';
import 'package:source_gen/src/annotation.dart';

T findAnnotation<T>(FieldElement field, Type outType) {
  var first = field.metadata
      .firstWhere((ann) => matchAnnotation(outType, ann), orElse: () => null);
  return first == null ? null : instantiateAnnotation(first);
}
```

Firstly,  `matchAnnotation` is not able to tell that `@hasMany` is an instance of `Relationship`. 
I hacked around this by instead searching for a `HasMany` or `HasOne`, etc.

But `instantiateAnnotation` crashes with the following error when encountering a child class:

```
PS C:\Users\thosa\Source\Angel\postgres> dart tool/build.dart
[INFO] Build: Reading cached dependency graph...
[INFO] Build: Reading cached dependency graph completed, took 15ms
[INFO] Build: Finalizing build setup...
[INFO] Build: Updating dependency graph with changes since last build.
[INFO] Build: Initializing inputs
[INFO] Build: Deleting previous outputs
[INFO] Build: Deleting 3 declared outputs which already existed on disk.
[INFO] Build: Finalizing build setup completed, took 192ms
[INFO] Build: Running build...
[INFO] runBuilder: Hm: make
[...]

Issue #117, opened 2017-01-07
Make StandaloneBuilder a separate class from GeneratorBuilder


natebosch:
The api is confusing as is since it only allows a single Generator in standalone mode. We should make that explicit through the signature.

Once the constructor for GeneratorBuilder doesn't have a throw we can also make it `const` which is a common pattern for builders.

natebosch:
Since we've got a good collection of use cases established I'm going to look at a deeper rewrite of the top level interfaces and how they can be put together. As part of that I'll rename `GeneratorBuilder` to `PartBuilder` and the new standalone concept will be called `SourceBuilder`.

Similarly `Generator` and `GeneratorForAnnotation` will be replace with some typedefs and I'll introduce `SourceGenerator` as the more general concept. We'll be able to put these concepts together to cover the 4 use cases we've seen.

- Build a `part of` file out of snippets for annotated code
- Build a `part of` file out of snippets for arbitrary `Element`s
- Build a standalone Dart source file for a library
- Build a standalone Dart source file + extra outputs for a library (with utilities to find annotated elements)

matanlurey:
@natebosch Want some help doing this, since I'm driving (some of the) requirements?

When you're back from leave ping me :)

natebosch:
I'll see if I can dig up the changes I started in this direction. I ended up hitting some places where I wanted to use functions but was forced to use classes because function generic arguments are erased at runtime - the UX wasn't what I hoped.

Adding this to my TODO list for when I return and I'll sync with you on either getting some help or putting out my prototype.

natebosch:
Should be pretty straightforward at this point. The decision we need to make is what public surface area we want: `PartBuilder` and `LibraryBuilder`, or `GeneratorBuilder` and `GeneratorBuilder.standalone`...

@kevmoo @matanlurey any opinion?

I think the Part / Library distinction is easier to understand, but GeneratorBuilder.standalone would be a smaller change since clients who aren't passing `isStandalone` today wouldn't need to change anything.

kevmoo:
Vote for LibraryBuilder + PartBuilder

CC @jakemac53

PR #159, opened 2017-06-19
Patch instantiateAnnotation to support extended annotation classes

thosakwe:
I've left details about an error I'm experiencing with source_gen in #155.

This PR will solve one of the two problems I noticed, which is that `instantiateAnnotation` does not work
for annotations that are extended from a parent class.

Please let me know if there is anything I should change before this PR is merged!

### DIFF
--- a/lib/src/annotation.dart
+++ b/lib/src/annotation.dart
@@ -151,6 +151,8 @@ dynamic _createFromConstructor(
       // set it as the argument value
 
       var initializer = ctor.constantInitializers.singleWhere((ci) {
+        // Avoid crashing on SuperConstructor
+        if (ci is SuperConstructorInvocation) return false;
         var expression = (ci as ConstructorFieldInitializer).expression;
         if (expression is SimpleIdentifier) {
           return expression.name == paramName;
--- a/lib/src/annotation.dart
+++ b/lib/src/annotation.dart
@@ -168,7 +168,7 @@ dynamic _createFromConstructor(
           // from the parent class.
 
           // 1. Find parent class
-          
+
 
           // 2. Determine the name of the field this parameter is assigned to on the parent class.
 
--- a/lib/src/annotation.dart
+++ b/lib/src/annotation.dart
@@ -150,6 +150,14 @@ dynamic _createFromConstructor(
       // field assigned in the object. Then we can take the field value and
       // set it as the argument value
 
+      // If the annotation class's constructor has no initializers and just
+      // forwards to a `super` class constructor, then there is no need
+      // to add a ctor argument name as a field value.
+      //
+      // In this case `singleWhere` would throw a StateError.
+      if (ctor.constantInitializers.length == 1 && ctor.constantInitializers.first is SuperConstructorInvocation)
+        continue;
+
       var initializer = ctor.constantInitializers.singleWhere((ci) {
         // Avoid crashing on SuperConstructor
         if (ci is SuperConstructorInvocation) return false;
--- a/lib/src/annotation.dart
+++ b/lib/src/annotation.dart
@@ -140,29 +140,41 @@ dynamic _createFromConstructor(
     ConstructorElementImpl ctor, DartObjectImpl obj) {
   var positionalArgs = [];
   var namedArgs = <Symbol, dynamic>{};
+  DartObjectImpl fieldObjectImpl;
+
   for (var p in ctor.parameters) {
     var paramName = p.name;
     String fieldName;
     if (p is FieldFormalParameterElement) {
       fieldName = p.name;
     } else {
+      // If the parameter is not a field parameter, then it is one of two things:
+      // * An argument corresponding to a field assigned in the object
+      // * An argument that is passed to a super constructor
+      //
+      // If the argument is one merely passed to a super constructor,
+      // then there will not be an entry in `obj.fields` for its name.
+      //
+      // So what should we add to positionalArgs/namedArgs instead???
+
+
       // Trying to find the relationship between the ctor argument name and the
       // field assigned in the object. Then we can take the field value and
       // set it as the argument value
 
-      if (ctor.constantInitializers.length == 1
-          && ctor.constantInitializers.first is SuperConstructorInvocation) {
-        // If the annotation class's constructor has no initializers and just
-        // forwards to a `super` class constructor, then there is no need
-        // to add a ctor argument name as a field value.
-        //
-        // In this case `singleWhere` would throw a StateError.
-        continue;
-      }
-
       var initializer = ctor.constantInitializers.singleWhere((ci) {
-        // Avoid crashing on SuperConstructor
-        if (ci is SuperConstructorInvocation) return false;
+        if (ci is SuperConstructorInvocation) {
+          // If this is a super constructor invocation, we need to find the right field
+          // from the parent class.
+
+          // 1. Find parent class
+          
+
+          // 2. Determine the name of the field this parameter is assigned to on the parent class.
+
+          // 3. Assign fieldObjectImpl
+        }
+
         var expression = (ci as ConstructorFieldInitializer).expression;
         if (expression is SimpleIdentifier) {
           return expression.name == paramName;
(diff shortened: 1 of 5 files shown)


lib/src/annotation.dart
```diff
@@ -150,7 +150,17 @@ dynamic _createFromConstructor(
       // field assigned in the object. Then we can take the field value and
       // set it as the argument value
 
+      // If the annotation class's constructor has no initializers and just
+      // forwards to a `super` class constructor, then there is no need
+      // to add a ctor argument name as a field value.
+      //
+      // In this case `singleWhere` would throw a StateError.
+      if (ctor.constantInitializers.length == 1 && ctor.constantInitializers.first is SuperConstructorInvocation)
```
jakemac53: nit: use curlies unless the whole statement fits on one line.

test/test_utils.dart
```diff
@@ -21,6 +22,10 @@ String getPackagePath() {
 
     _packagePathCache = p.normalize(p.join(p.dirname(currentFilePath), '..'));
   }
+
+  // On Windows, the tests will crash because _packagePathCache starts with a backslash.
+  if (Platform.isWindows && _packagePathCache.startsWith('\\'))
```
jakemac53: nit: move this into the if so it doesn't run every time this function is called.

lib/src/annotation.dart
```diff
@@ -150,7 +150,17 @@ dynamic _createFromConstructor(
...
+        continue;
```
jakemac53: I don't quite understand under which scenario this happens, in the test provided there are zero parameters so we wouldn't reach this code. If you could add a test outlining the case that this is handling that would help clarify things :).

It also doesn't seem like skipping a parameter entirely is ever the right thing to do?

test/test_files/annotations.dart
```diff
@@ -56,6 +56,10 @@ class PublicAnnotationClass {
         child2 = const PublicAnnotationClass.withAnIntAsOne();
 }
 
+class ExtendedAnnotationClass extends PublicAnnotationClass {
+  const ExtendedAnnotationClass():super();
```
jakemac53: What happens if we have a parameter that is forwarded to the super constructor? If that isn't supported we should throw an error calling it out explicitly I think.

lib/src/annotation.dart
thosakwe: This happens in a case like the following:

```dart
// Relationship is also an annotation class
class HasMany extends Relationship {
 // This `super` call 
 const HasMany():super();
}
```

`ctor.constructorInitializers` will contain a `SuperConstructorInvocation`, which throws an error because we try to cast it as `ConstructorFieldInitializer`.

I added a test involving this, but perhaps I wasn't clear about which it was:
```dart
test('annotated with extended class', () {
      var instance = _getInstantiatedAnnotation(libElement, 'WithChildClassAnnotation');
      expect(instance is defs.PublicAnnotationClass, isTrue);
      expect(instance is defs.ExtendedAnnotationClass, isTrue);
      expect(instance.anInt, 0);
      expect(instance.aBool, isFalse);
    });

// annotated_classes.dart
@ExtendedAnnotationClass()
class WithChildClassAnnotation {}

// annotations.dart
class ExtendedAnnotationClass extends PublicAnnotationClass {
  const ExtendedAnnotationClass():super();
}
```

Also, yes, it seems kind of iffy to skip a parameter, but it would always throw an error because the code would cast each member of `constructorInitalizers` to `ConstructorFieldInitializer`, which would always throw a type error. I'm hoping there is a better way to do this, but this seems to work.

lib/src/annotation.dart
thosakwe: Ok. Definitely will fix this

test/test_files/annotations.dart
thosakwe: I'll have to add a test case for that; didn't think about that. I'm thinking that it will work as-is, but of course, no way to be sure without testing it.

test/test_utils.dart
thosakwe: Will do.

lib/src/annotation.dart
jakemac53: > This happens in a case like the following:
> ```dart
> // Relationship is also an annotation class
> class HasMany extends Relationship {
>  // This `super` call 
>  const HasMany():super();
> }
>```

Why I am confused is because the constructor in the test/example doesn't have any parameters, and this is inside a for loop that is looping over the parameters, so I wouldn't expect it to get executed at all?

> Also, yes, it seems kind of iffy to skip a parameter, but it would always throw an error because the code would cast each member of constructorInitalizers to ConstructorFieldInitializer, which would always throw a type error. I'm hoping there is a better way to do this, but this seems to work.

I don't think we have to support the following, but I want to make sure the error is reasonable:

```dart
class A {
  final String foo;
  const A(this.foo);
}

class B extends A {
 const B(String foo) : super(foo);
}
```

I think today you would just end up bailing out and then not recreating the annotation properly later on, which would be very hard to track down.

This is essentially the same thing as what I wanted to have a test added for :).

thosakwe:
Okay, so I just pushed a commit, but it's not ready yet. I realized that the approach of `_createFromConstructor` assumes that any constructor parameter that is not a `this.x`-type is a constant initializer (i.e. `MyType(): foo = 'bar', baz = 'quux'`).

So if you have a parameter that is forwarded to a super constructor, a state error will be thrown, because the code will try to resolve a field on the object that might not exist.

So in a case like this, I think we'd have to search the parent type for the correct name of the field that the parameter is ultimately assigned to:

```dart
class A {
 final int n;
 const A(this.n);
}

class B extends A {
 const B(int forwardedN):super(forwardedN);
}

// Would have to somehow find out that `forwardedN` becomes the `n` field...
```

But this also has a problem: it would only work if the parameter is used by the parent class. If there were a `C` annotation class that also took a `forwardedN` and passed it up all the way to `A`, there would be an error.

So, right now, I'm not sure how to properly patch this.

thosakwe:
Is there any way to resolve all the positional+named args passed to a constructor, and then just initialize from that? 

```dart
@A(2, 3, '4', five: 5)

// This would produce:
positional: [2, 3, '4']
named: {'five': 5}
```

The current way of assuming that all parameters directly map to a field on the annotation class is what's preventing it from working with extended classes, I believe.

jakemac53:
cc @bwilkerson for guidance. I have a feeling all this code could/should be refactored as the analyzer apis have changed a lot and there may be better support now.

thosakwe:
Okay, cool. Fingers crossed.

kevmoo:
@jakemac53 @matanlurey what do we want to do here?

matanlurey:
My general feeling is the ability to recreate a class from an annotation has enough corner cases where I'm not sure this is enough, even if we get it fixed. Personally I'd prefer if `GeneratorForAnnotation` had following API instead:

```dart
abstract class GeneratorForAnnotation<T, E extends Element> implements Generator {
  const GeneratorForAnnotation();

  @protected
  T createAnnotation(DartObject object) => _useMirrorsToCreate(T, object);

  Future<String> generateForAnnotation(E element, T annotation);
}
```

Then at least users could override `createAnnotation` to cover their own corner cases.

natebosch:
@matanlurey - perhaps we should avoid `createAnnotation` anywhere in `GeneratorForAnnotation`? With the new utilities you've added do you think it would be a usability concern to implement as `Future<String> generateForAnnotatedElement(E element, DartObject annotation)`?

The annotation matching could happen with the new type utilities and I don't think we need an instance of `T` for any other reason than usability.

Having an instance of `T` is definitely nicer than having a `DartObject` - so for the cases that can be handled we could expose a utility and authors that want it can take that step themselves.

We could wrap that API change up with the other breaking changes in #117

matanlurey:
It seems like a less breaking change to at least keep current semantics for now, though I agree longer-term we should try not to do anything too magical for users if we don't have predictable behavior.

kevmoo:
Where are we here @thosakwe @matanlurey @natebosch ?

matanlurey:
I don't have an opinion on this change, since I don't even like it existing to begin with :)

thosakwe:
Regardless of this specific PR being merged, it would definitely be nice to be able to use source_gen with annotation classes that extend from other classes.

natebosch:
I think we're going to move towards using `ConstantReader`. See #224 

ConstantReader handles reading fields up a chain of superclasses so that should resolve this concern.

thosakwe:
Sounds good; I've been using `ConstantReader` in my project. Thanks for the responses.

Issue #161, opened 2017-06-19
Add 'package:'-level types to type_checker_test


matanlurey:
Specifically after a number of `build_test` features are added:

- https://github.com/dart-lang/build/issues/297
- https://github.com/dart-lang/build/issues/298
- https://github.com/dart-lang/build/issues/299

Issue #160, opened 2017-06-19
Add 'findType', like 'LibraryElement.getType' but...


matanlurey:
... includes walking `export` directives.

Open for alternative names, or closing this bug if there is something I'm missing.

natebosch:
I like `findType(LibraryElement, String)` as a utility method in this package.

We could consider adding it to Resolver directly - but I like the idea of keeping that class narrow and building up utils here.

matanlurey:
Me too. Based on my new definition of `source_gen`, I'd rather keep `build` to be super-small and stable long-term, while new techniques and utilities go here instead. I'll send a PR.

PR #165, opened 2017-06-20
Add findType, and fix a bug in TypeChecker.

matanlurey:
Closes https://github.com/dart-lang/source_gen/issues/160.

Also added test cases, which in turn found a bug. Whad'ya know @jakemac53 🐑 

Closes https://github.com/dart-lang/source_gen/issues/161.

### DIFF
--- a/lib/src/type_checker.dart
+++ b/lib/src/type_checker.dart
@@ -28,6 +28,12 @@ abstract class TypeChecker {
   ///   'dart:collection#LinkedHashMap',
   /// );
   /// ```
+  ///
+  /// **NOTE**: This is considered a more _brittle_ way of determining the type
+  /// because it relies on knowing the _absolute_ path (i.e. after resolved
+  /// `export` directives). You should ideally only use `fromUrl` when you know
+  /// the full path (likely you own/control the package) or it is in a stable
+  /// package like in the `dart:` SDK.
   const factory TypeChecker.fromUrl(dynamic url) = _UriTypeChecker;
 
   /// Returns the first constant annotating [element] that is this type.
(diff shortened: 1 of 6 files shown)


lib/src/type_checker.dart
```diff
@@ -28,6 +28,12 @@ abstract class TypeChecker {
   ///   'dart:collection#LinkedHashMap',
   /// );
   /// ```
+  ///
+  /// **NOTE**: This is considered a more _brittle_ way of determining the type
+  /// because it relies on knowing the _absolute_ path (i.e. after resolved
+  /// `export` directives). You should ideally only use `fromUrl` when you know
+  /// the full path (likely you own/control the package) or it is in a stable
+  /// package like in the `dart:` SDK.
```
kevmoo: Maybe "stable package or the Dart SDK via `dart:` URLs.

test/find_type_test.dart
```diff
@@ -0,0 +1,43 @@
...
+
+import 'package:analyzer/dart/element/element.dart';
+import 'package:build_test/build_test.dart';
+import 'package:source_gen/source_gen.dart';
+import 'package:test/test.dart';
+
+void main() {
+  LibraryElement library;
+
+  setUpAll(() async {
```
jakemac53: Just curious, I notice you use this `setUpAll` a lot but I don't understand what the value is over just regular setup in the top of `main`?

matanlurey: `setUpAll` is only invoked once, not once per `test`.

The body of this function (doing analysis) is expensive, and doesn't change between tests, so this makes the test take ~20s instead of minutes.

jakemac53: `main` is only invoked once as well :).

I talked with @kevmoo and I guess you get better errors if the setup fails at least so there is that advantage. Also if all tests end up skipped then it wouldn't execute this code.

matanlurey: Yeah, also it makes the use of `async` a little claner.

Issue #169, opened 2017-06-20
Add helpers for interacting with constants/DartObject


matanlurey:
In `AngularDart` @alorenzen produced a number of `coerceX` methods (i.e. `coerceBool`), because we quickly found out that `DartObject#getField` rarely returns what you want - it only reads a constant fields _directly_ from the class (not delegating to `super.`).

I'd like to add them to this library, but not sure of the form. Some options:

* As-is (`coerceBool`, `coerceInt`, so-on)
* With a more descriptive name - i.e. `readFieldBool`
* Some sort of const-object wrapper, i.e.

```dart
abstract class Constant {
  factory Constant(DartObject obj) => ...

  bool get isBool;

  bool readBool({bool defaultTo()};
}
```

Thoughts? @jakemac53 @natebosch @kevmoo @alorenzen

jakemac53:
We could also add to this class helper methods to get an actual instance (only throws if you are not importing the required types), or get as a String the code to recreate the constant?

matanlurey:
Definitely! I'm going to take this as a "sure, let's do it". 

Do you have a preference on class wrapper versus top-level methods?

jakemac53:
I think a wrapper makes the most sense.

alorenzen:
+1 to class wrapper

PR #171, opened 2017-06-20
Add a Constant helper class.

matanlurey:
Partial progress towards https://github.com/dart-lang/source_gen/issues/169.

`List`, `Map`, and other types are a little trickier, so hoping to get some incremental work first.

### DIFF
--- a/lib/src/constants.dart
+++ b/lib/src/constants.dart
@@ -11,6 +11,8 @@ Null _alwaysNull() => null;
 bool _isNull(DartObject object) => object?.isNull != false;
 
 /// Similar to [DartObject.getField], but traverses super classes.
+///
+/// Returns `null` if ultimately [field] is never found.
 DartObject _getFieldRecursive(DartObject object, String field) {
   if (_isNull(object)) {
     return null;
@@ -31,18 +33,12 @@ abstract class Constant {
       _isNull(object) ? const _NullConstant() : new _Constant(object);
 
   /// Returns whether this constant represents a `bool` literal.
-  ///
-  /// If `true`, [boolValue] will return either `true` or `false` (not throw).
   bool get isBool;
 
   /// Returns this constant as a `bool` value.
-  ///
-  /// Throws [FormatException] if [isBool] is `false`.
   bool get boolValue;
 
   /// Returns whether this constant represents an `int` literal.
-  ///
-  /// If `true`, [intValue] will return an `int` (not throw).
   bool get isInt;
 
   /// Returns this constant as an `int` value.
@@ -148,7 +144,7 @@ class _Constant implements Constant {
   bool get isInt => _object.toIntValue() != null;
 
   @override
-  bool get isNull => false;
+  bool get isNull => _isNull(_object);
 
   @override
   bool get isString => _object.toStringValue() != null;
--- a/lib/src/constants.dart
+++ b/lib/src/constants.dart
@@ -4,9 +4,6 @@
 
 import 'package:analyzer/dart/constant/value.dart';
 
-/// Always returns `null` (used as a default for `defaultTo` methods).
-Null _alwaysNull() => null;
-
 /// Returns whether or not [object] is or represents a `null` value.
 bool _isNull(DartObject object) => object?.isNull != false;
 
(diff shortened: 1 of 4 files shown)


lib/src/constants.dart
```diff
@@ -0,0 +1,171 @@
...
+///
+/// Unlike [DartObject.getField], all `readX` methods attempt to access super
+/// classes for the field value if not found.
+abstract class Constant {
+  factory Constant(DartObject object) =>
+      _isNull(object) ? const _NullConstant() : new _Constant(object);
+
+  /// Returns whether this constant represents a `bool` literal.
+  ///
+  /// If `true`, [boolValue] will return either `true` or `false` (not throw).
```
jakemac53: I don't think you need these lines on the is* getters.

lib/src/constants.dart
```diff
@@ -0,0 +1,171 @@
...
+  /// Returns whether this constant represents `null`.
+  bool get isNull;
+
+  /// Reads[ field] from the constant as another constant value.
+  Constant read(String field);
+
+  /// Reads [field] from the constant as a boolean.
+  ///
+  /// If the resulting value is `null`, uses [defaultTo] if defined.
+  bool readBool(String field, {bool defaultTo()});
```
jakemac53: I am a bit confused why these methods exist in addition to the *Value getters?

lib/src/constants.dart
```diff
@@ -0,0 +1,171 @@
...
+      : throw new FormatException('Not a String', _object);
+
+  @override
+  bool get isBool => _object.toBoolValue() != null;
+
+  @override
+  bool get isInt => _object.toIntValue() != null;
+
+  @override
+  bool get isNull => false;
```
jakemac53: For future proofing it might be best to do `_isNull(_object)` instead?

test/constants_test.dart
```diff
@@ -0,0 +1,91 @@
...
+import 'package:build_test/build_test.dart';
+import 'package:source_gen/source_gen.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('Constant', () {
+    List<Constant> constants;
+
+    setUpAll(() async {
+      constants = (await resolveSource(r'''
```
jakemac53: nit: assign the result of `resolveSource` to a local variable to make this a bit easier to follow. In general I try not to wrap await calls in parens and use local vars instead because it becomes hard to follow.

lib/src/constants.dart
```diff
@@ -0,0 +1,171 @@
...
+
+import 'package:analyzer/dart/constant/value.dart';
+
+/// Always returns `null` (used as a default for `defaultTo` methods).
+Null _alwaysNull() => null;
+
+/// Returns whether or not [object] is or represents a `null` value.
+bool _isNull(DartObject object) => object?.isNull != false;
+
+/// Similar to [DartObject.getField], but traverses super classes.
```
matanlurey: Done.

lib/src/constants.dart
```diff
@@ -0,0 +1,171 @@
...
+DartObject _getFieldRecursive(DartObject object, String field) {
+  if (_isNull(object)) {
+    return null;
```
matanlurey: If you don't mind I'm addressing that in the next PR - the code to check all fields is a little meaty.

Filed https://github.com/dart-lang/source_gen/issues/173.

lib/src/constants.dart
matanlurey: Done.

lib/src/constants.dart
matanlurey: A DartObject can either _be_ a literal or _have_ fields (that are either literals or objects themselves).

Because `Constant` wraps a `DartObject`, it makes sense to make this easier. It also enables (eventually) added support for using a dot notation, or something like:

```dart
constant.readBool('from.another.thing');
```

If you think it's not worth it and we should just have `xValue`, happy to revert though.

lib/src/constants.dart
matanlurey: I think it's impossible to get into a state where that is true, but no harm doing this.

test/constants_test.dart
matanlurey: Done.

lib/src/constants.dart
matanlurey: On second thought I agree with you, this makes `defaultTo` not necessary, for example.

matanlurey:
@kevmoo @jakemac53 Let me know if you'd rather this be called `ConstantReader` (in-line with `LibraryReader` in https://github.com/dart-lang/source_gen/pull/172.

kevmoo:
`ConstantReader` – sgtm – @jakemac53 ?

Issue #170, opened 2017-06-20
Create a helper for a LibraryElement (Library)


matanlurey:
Similar to https://github.com/dart-lang/source_gen/issues/169, it actually might be better to move `findType` to be a method of a helper called `Library` that wraps `LibraryElement`, versus polluting a bunch of top-level methods.

WDUT @jakemac53?

jakemac53:
It sounds logical, but I do fear a bit that we are going to fall into a trap of creating wrappers for all core concepts now hahaha (Class, Directive, etc etc etc)

matanlurey:
I think we can push back from most other requests, `LibraryElement` and `DartObject` are the two places I know of that are pretty complicated to use correctly while doing static analysis. We already have a `ClassElement` wrapper called `TypeChecker` ;)

PR #172, opened 2017-06-20
Convert findType -> Library#findType.

matanlurey:
Closes https://github.com/dart-lang/source_gen/issues/170.


lib/src/library.dart
```diff
@@ -0,0 +1,25 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/src/dart/resolver/scope.dart';
+
+/// A high-level wrapper API with common functionality for [LibraryElement].
+class Library {
+  final LibraryElement _element;
+
+  Namespace _namespace;
```
jakemac53: Consider making this a getter which lazily creates the namespace, which would clean up the `findType` code a fair bit I think.

lib/src/library.dart
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/src/dart/resolver/scope.dart';
+
+/// A high-level wrapper API with common functionality for [LibraryElement].
+class Library {
```
kevmoo: The name bothers me a bit. Seems a bit too generic.

Maybe LibraryHelper? Wrapper?

Not a huge deal...just thinking...

lib/src/library.dart
matanlurey: Done.

lib/src/library.dart
matanlurey: Renamed to `LibraryReader`.

Issue #157, opened 2017-06-18
Support non-library part files


matanlurey:
i.e. the format of `part of package:foo/foo.dart`.

lrhn:
Can you elaborate on what you want to change, and to what?

natebosch:
To be clear - we're talking about outputting that format?

Definitely should, we'll need to make sure we're ready for the SDK constraint to get that high.

cc @kevmoo - any reason not to do this now?

kevmoo:
It'll be supported officially in 1.25 - https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md

Surprised that `package:` support is allowed for `part of`. Really, @matanlurey ?

matanlurey:
I'm not sure, maybe it is just relative URLs.

@natebosch @kevmoo See https://github.com/dart-lang/source_gen/pull/158, which adds a way to do this in a backwards-compatible way. I don't think there is a way to auto-detect your SDK version though.

kevmoo:
...and it's the dance between the version of the SDK running the generator, the target SDK for the package into which the code is being generated...and the version pkg/analyzer running behind source_gen.

Fun!

matanlurey:
I suspect we should make this a flag on the `GeneratorBuilder` @kevmoo:

```dart
class GeneratorBuilder {
  GeneratorBuilder(this.generators, {bool requireLibraryDirective: true, ...})
}
```

Users that want to set a constraint of `>=1.25.0` can set this flag to `false`.

We can default it to `true` in a future release. WDUT?

kevmoo:
SGTM – @jakemac53

PR #179, opened 2017-06-21
GeneratorBuilder flag for "part of" without "library".

matanlurey:
Closes https://github.com/dart-lang/source_gen/issues/157.
Closes https://github.com/dart-lang/source_gen/pull/138.

This approach will not lock-in all future users of `source_gen` to an unreleased `1.25.0`. We can flip the default value in a future (breaking) release.

/cc @davidmorgan

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,12 @@
   ```
   
   In Dart SDK `>=1.25.0` this can be relaxed as `part of` can refer to a path.
+  To opt-in, `GeneratorBuilder` now has a new flag, `requireLibraryDirective`.
+  Set it to `false`, and also set your `sdk` constraint appropriately:
+  
+  ```yaml
+    sdk: '>=1.25.0 <2.0.0'
+  ```
 
 * Added `findType`, an utility method for `LibraryElement#getType` that also
   traverses `export` directives for publicly exported types. For example, to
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,12 @@
   ```
   
   In Dart SDK `>=1.25.0` this can be relaxed as `part of` can refer to a path.
+  To opt-in, `GeneratorBuilder` now has a new flag, `requireLibraryDirective`.
```
jakemac53: Just an idea, but you can check the SDK version using `Platform.version` from `dart:io`, and we could make the default be based on that even? That would also likely require a dependency on `pub_semver`  but that should be ok.

matanlurey: Fwiw I actually like this idea, but it hasn't technically landed yet even at -dev. Maybe we can set the default _based_ on that once 1.25.0 lands officially?

jakemac53: Sounds good - if it hasn't landed officially in any release there is no sdk constraint we can set either right.... so we can't really release this feature at all until then.

matanlurey: I figured adding this flags gives users the chance to make that choice for themselves versus waiting for us to do the right thing ;)

jakemac53: Also I forgot about the fact that you need to update your sdk constraint if you are publishing these files as part of a package, so using it automatically is a no-go.

What we could do though is just add the sdk constraint in this package?

jakemac53: But ya in any case adding the flag seems like an easy workaround to sidestep the issue

Issue #173, opened 2017-06-20
Constant: Reading a field that isn't defined should throw.


matanlurey:
Relevant after merging https://github.com/dart-lang/source_gen/pull/171.

matanlurey:
Started working on this @ https://github.com/matanlurey/source_gen/tree/constant-more.

Will send a PR after #171 closes.

PR #180, opened 2017-06-22
More features for ConstantReader

matanlurey:
Closes https://github.com/dart-lang/source_gen/issues/173

* Reading a field that doesn't exist now throws `FormatException`

Closes https://github.com/dart-lang/source_gen/issues/169

* Now supports `listValue` and `mapValue`


kevmoo:
@matanlurey appease the travis gods...

pubspec.yaml
```diff
@@ -8,12 +8,12 @@ environment:
 dependencies:
   analyzer: '>=0.29.2 <0.31.0'
   build: ^0.9.0
+  collection: ^1.1.2
```
kevmoo: was this an oops from a previous commit?

lib/src/constants.dart
```diff
@@ -3,6 +3,25 @@
...
 
 import 'package:analyzer/dart/constant/value.dart';
+import 'package:analyzer/dart/element/element.dart';
+
+/// Throws an exception if [root] or its super(s) does not contain [name].
+void _assertHasField(ClassElement root, String name) {
+  var element = root;
+  while (element != null) {
+    final field = element.getField(name);
+    if (field != null) {
```
natebosch: [nit] consider collapsing this to 1 line so it doesn't take more visual weight than it earns.

lib/src/constants.dart
```diff
@@ -3,6 +3,25 @@
...
+      return;
+    }
+    element = element.supertype?.element;
+  }
+  final allFields = root.fields.toSet();
+  root.allSupertypes.forEach((t) => allFields.addAll(t.element.fields));
```
natebosch: [nit] https://www.dartlang.org/guides/language/effective-dart/usage#avoid-using-iterableforeach-with-a-function-literal

Could use a for loop, or I think it might still fit on 1 line to split this into 2 steps.

`root.allSupertype.map((t) => t.element.fields).forEach(allFields.addAll);`

or

`allFields.addAll(root.allSupertypes.expand((t) => t.element.fields));`

lib/src/constants.dart
```diff
@@ -43,6 +62,26 @@ abstract class ConstantReader {
   /// Throws [FormatException] if [isInt] is `false`.
   int get intValue;
 
+  /// Returns whether this constant represents a `List` literal.
```
natebosch: [nit] `get` methods should be documented like a variable or a field - with a noun so you can omit 'Returns'.

Similar below.

matanlurey:
@natebosch Oops. I just saw this after refreshing. I'll fix comments upstream.

Issue #178, opened 2017-06-21
API for "reviving" `DartObject`


matanlurey:
One of the hardest and most brittle parts of the AngularDart compiler...

Basically, we need to go from a computed `DartObject` to some sort of model class that enables code generation of an *identical* object. Here are some simple (trivial) examples that are solvable with `ConstantReader` today:

* `String`
* `int` / `double`
* `bool`
* `null`
* `List<...>` or `Map<...>` of these types, recursively

However, it's much harder with classes and enums. Consider these:

* `const Duration(seconds: 5)`
* `const SomeClass._privateConstructor()`
* `Enum.someValue`

---

We _could_ do this with `ConstantReader`, if we add a few more methods:

```dart
abstract class ConstantReader {
  bool get isEnum;
  RevivableEnum get enumValue;

  bool get isInstance;
  RevivableInstance get instanceValue;
}

abstract class RevivableEnum implements ConstantReader {
  Uri get uri;
  String get name;
  int get index;
}

abstract class RevivableInstance implements ConstantReader {
  Uri get uri;
  String get name;
  String get constructor;
  List<ConstantReader> get positionalArguments;
  Map<String, ConstantReader> get namedArguments;
}
```

We could also do error handling, i.e., a flag for checking if it is revivable (visible):

```dart
abstract class RevivableInstance implements ConstantReader {
  ...
  bool get isVisible;
}
```

This isn't a perfect API, but it is probably better for some of this to live in `source_gen` (and have closer watch by the ecosystem team and analysis team) than buried in the AngularDart compiler.

/cc @jakemac53 @alorenzen @kevmoo

jakemac53:
I am not 100% sure about the `Revivable*` names but generally this SGTM.

We probably also want to probably add a base class, something similar to this:

```dart
abstract RevivableBase<T> {
  String toCodeString();
  T toInstance();
}
```

matanlurey:
I mostly agree - except the `toCodeString` - this seems too difficult to get right for all users, and then we are sort of duplicating logic for from `code_builder`.

matanlurey:
I'm OK with the progress here so far.

bitsydarel:
Hi guys, so what's the current api to get the value of a enum using constantReader ?

bitsydarel:
is their a toInstance ?

natebosch:
> what's the current api to get the value of a enum using constantReader ?

I don't think we have one. We could review a contribution adding an API to `ConstantReader` for this.

[...]

PR #182, opened 2017-06-22
Constants reviver

matanlurey:
Partial support of https://github.com/dart-lang/source_gen/issues/178.

Still missing as of this PR:
* Any sort of sugar for `enum`, though it should work out of the box anyway.
* Error recovery: We just return `null` when something is not resolvable, ideally I'd like to either throw or return a specialized instance (`Unrevivable`) that explains why not (usually due to a private constructor, or not being able to trace how the object was created

Good starting point though I think.

### DIFF
--- a/lib/src/constants.dart
+++ b/lib/src/constants.dart
@@ -5,6 +5,8 @@
 import 'package:analyzer/dart/constant/value.dart';
 import 'package:analyzer/dart/element/element.dart';
 
+import 'type_checker.dart';
+
 /// Throws an exception if [root] or its super(s) does not contain [name].
 void _assertHasField(ClassElement root, String name) {
   var element = root;
@@ -95,6 +97,9 @@ abstract class ConstantReader {
   /// Returns whether this constant represents `null`.
   bool get isNull;
 
+  /// Returns whether this constant matches [checker].
+  bool instanceOf(TypeChecker checker);
+
   /// Reads[ field] from the constant as another constant value.
   ConstantReader read(String field);
 }
@@ -141,6 +146,9 @@ class _NullConstant implements ConstantReader {
   bool get isString => false;
 
   @override
+  bool instanceOf(TypeChecker checker) => false;
+
+  @override
   ConstantReader read(_) => throw new UnsupportedError('Null');
 }
 
@@ -176,16 +184,20 @@ class _Constant implements ConstantReader {
   bool get isInt => _object.toIntValue() != null;
 
   @override
-  bool get isList => _object?.toListValue() != null;
+  bool get isList => _object.toListValue() != null;
 
   @override
   bool get isNull => _isNull(_object);
 
   @override
-  bool get isMap => _object?.toMapValue() != null;
+  bool get isMap => _object.toMapValue() != null;
 
   @override
   bool get isString => _object.toStringValue() != null;
+
+  @override
+  bool instanceOf(TypeChecker checker) =>
+      checker.isAssignableFromType(_object.type);
 
   @override
   ConstantReader read(String field) {
--- a/lib/src/constants.dart
+++ b/lib/src/constants.dart
@@ -5,6 +5,7 @@
 import 'package:analyzer/dart/constant/value.dart';
 import 'package:analyzer/dart/element/element.dart';
 
+import 'revive.dart';
 import 'type_checker.dart';
 
 /// Throws an exception if [root] or its super(s) does not contain [name].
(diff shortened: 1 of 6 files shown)


lib/src/revive.dart
```diff
@@ -0,0 +1,79 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/constant/value.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/src/dart/constant/value.dart';
+
+import 'utils.dart';
+
+/// Base interface for a meta-analyzed [DartObject].
+abstract class Revivable {}
```
kevmoo: Weird to have an empty base class w/ no members – and one implementation?

test/constants_test.dart
```diff
@@ -128,4 +128,52 @@ void main() {
       expect($deprecated.instanceOf(check), isTrue, reason: '$deprecated');
     });
   });
+
+  group('Reviable', () {
```
kevmoo: missing a `v`

lib/src/revive.dart
```diff
@@ -0,0 +1,79 @@
...
+FieldElement _findPublicConstField(DartObject object, ClassElement clazz) =>
+    clazz.fields.firstWhere(
+        (f) => f.isConst && f.isPublic && f.computeConstantValue() == object,
+        orElse: () => null);
+
+ConstructorInvocation _findPublicConstConstructor(
+  DartObject object,
+  ClassElement clazz,
+) {
+  final invocation = (object as DartObjectImpl).getInvocation();
```
kevmoo: `getInvocation` may return null – need to handle that!

matanlurey:
@kevmoo PTAL, lots of changes.

lib/src/revive.dart
matanlurey: Obsolete.

lib/src/revive.dart
```diff
@@ -0,0 +1,79 @@
...
+import 'package:analyzer/src/dart/constant/value.dart';
+
+import 'utils.dart';
+
+/// Base interface for a meta-analyzed [DartObject].
+abstract class Revivable {}
+
+/// Returns a revivable instance of [object].
+///
+/// Optionally specify the [clazz] type that contains the constructor.
```
matanlurey: Obsolete.

lib/src/constants.dart
```diff
@@ -151,6 +176,14 @@ class _Constant implements ConstantReader {
...
 
   @override
+  dynamic get anyValue =>
+      _object.toBoolValue() ??
+      _object.toIntValue() ??
+      _object.toStringValue() ??
+      _object.toListValue() ??
+      _object.toMapValue();
+
+  @override
```
kevmoo: You're missing `symbol` here – it's supported on `DartObject`

You'll just want to wrap it in `new Symbol` if it returns a String.

kevmoo: Also missing `double`

lib/src/revive.dart
```diff
@@ -0,0 +1,80 @@
...
+/// Returns a revivable instance of [object].
+///
+/// Optionally specify the [library] type that contains the reference.
+///
+/// Returns [null] if not revivable.
+Revivable reviveInstance(DartObject object, [LibraryElement library]) {
+  library ??= object.type.element.library;
+  var url = Uri.parse(urlOfElement(object.type.element));
+  final clazz = object?.type?.element as ClassElement;
+  for (final e in clazz.fields.where(
```
natebosch: why does this need to be a for loop if the first line is a return? Is this a typical pattern to handle the 0 or first case?

would this be more readable as

```
var matchingFields = clazz.fields.where((f) => ...);
if (matchingFields.isNotEmpty) {
return new Revivable._(source: url, accessor: matchinfFields.first.name);
}

File: lib/src/revive.dart [hunk 7]
```diff
@@ -0,0 +1,80 @@
...
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/src/dart/constant/value.dart';
+
+import 'utils.dart';
+
+/// Returns a revivable instance of [object].
+///
+/// Optionally specify the [library] type that contains the reference.
+///
+/// Returns [null] if not revivable.
```
natebosch: Do we have known cases we know won't work? Could we expand this?

lib/src/revive.dart
kevmoo: Or just `firstWhere` – or even better – `singeWhere` – assuming more than one would be a bug

PR #183, opened 2017-06-22
isAssignable, isSuper fix

kevmoo:


kevmoo:
@matanlurey – fixed it. PTAL

PR #184, opened 2017-06-23
Add types to ConstantReader, fix issue with parts, migrate annotation helpers

kevmoo:


lib/src/annotation.dart
```diff
@@ -2,24 +2,22 @@
...
-import 'package:path/path.dart' as p;
+
+import 'constants.dart';
+import 'type_checker.dart';
 
 dynamic instantiateAnnotation(ElementAnnotation annotation) {
   var annotationObject = annotation.constantValue;
   try {
-    return _getValue(annotationObject, annotation.element.context.typeProvider);
+    return _getValue(annotation.constantValue);
```
matanlurey: AFAIK you always want to use `computeConstantValue`.

lib/src/constants.dart
```diff
@@ -98,6 +99,36 @@ abstract class ConstantReader {
...
+
+  /// Returns whether this constant represents a `Type` literal.
+  ///
+  /// If `true`, [typeValue] will return a `DartType` (not throw).
+  bool get isType;
+
+  /// Returns a [DartType] representing this as a `Type` value.
+  ///
+  /// Throws [FormatException] if [isType] is `false`.
+  DartType get typeValue;
```
kevmoo: @matanlurey – can't get the actual `Type` because runtime vs static.

lib/src/annotation.dart
```diff
@@ -51,33 +49,35 @@ dynamic instantiateAnnotation(ElementAnnotation annotation) {
       "${valueDeclaration.runtimeType}.");
 }
 
-dynamic _getValue(DartObject object, TypeProvider typeProvider) {
-  if (object.isNull) {
+dynamic _getValue(DartObject object) {
+  var reader = new ConstantReader(object);
+
+  if (reader.isNull) {
```
matanlurey: Optional: You can use `anyValue` for this purpose if you'd like.

lib/src/constants.dart
```diff
@@ -180,6 +228,9 @@ class _Constant implements ConstantReader {
       _object.toBoolValue() ??
       _object.toIntValue() ??
       _object.toStringValue() ??
+      _object.toDoubleValue() ??
+      (isSymbol ? this.symbolValue : null) ??
```
kevmoo: Special case here to return an actual `Symbol` and not a String

lib/src/constants.dart
```diff
@@ -221,6 +272,28 @@ class _Constant implements ConstantReader {
...
+  Symbol get symbolValue => isSymbol
+      ? new Symbol(_object.toSymbolValue())
+      : _throw('Symbol', _object);
+
+  @override
+  bool get isType => _object.toTypeValue() != null;
+
+  @override
+  DartType get typeValue =>
+      isType ? _object.toTypeValue() : _throw("Type", _object);
```
matanlurey: Consistent use of ' please.

lib/src/constants.dart
```diff
@@ -160,6 +190,24 @@ class _NullConstant implements ConstantReader {
   bool get isString => false;
 
   @override
+  bool get isDouble => false;
+
+  @override
+  double get doubleValue => _throw("double");
```
matanlurey: '

lib/src/constants.dart
```diff
@@ -160,6 +190,24 @@ class _NullConstant implements ConstantReader {
...
+
+  @override
+  bool get isSymbol => false;
+
+  @override
+  Symbol get symbolValue => _throw("Symbol");
```
matanlurey: '

lib/src/constants.dart
```diff
@@ -160,6 +190,24 @@ class _NullConstant implements ConstantReader {
...
+
+  @override
+  get isType => false;
+
+  @override
+  DartType get typeValue => _throw("Type");
```
matanlurey: '

lib/src/constants.dart
kevmoo: Huh?

kevmoo: Oh...geez...

lib/src/annotation.dart
kevmoo: Sadly no, we need to talk about that.

Type, List, and Map are weird – you get DartType.

I think we should update `anyValue` to exclude those and explain why.

lib/src/constants.dart
kevmoo: done

lib/src/constants.dart
kevmoo: done

lib/src/constants.dart
kevmoo: done

lib/src/constants.dart
kevmoo: done – enable a lint 😁

lib/src/annotation.dart
kevmoo: Just ran a little test – in all cases `constantValue` and `computeConstantValue` return the same result. 🤷‍♂️

PR #185, opened 2017-06-23
Use revivable for annotation helpers

kevmoo:


kevmoo:
@matanlurey – look at all of that deleted code. Reuse FTW

PR #191, opened 2017-06-27
Patch internal cases, bump.

matanlurey:
The changes in `builder.dart` specifically were required to land internally.

### DIFF
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -60,11 +60,9 @@ class GeneratorBuilder extends Builder {
     if (generatedExtension == null) {
       throw new ArgumentError.notNull('generatedExtension');
     }
-    if (generatedExtension.isEmpty ||
-        !generatedExtension.startsWith('.') ||
-        !generatedExtension.endsWith('.dart')) {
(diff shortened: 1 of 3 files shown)


lib/src/builder.dart
```diff
@@ -94,8 +92,14 @@ class GeneratorBuilder extends Builder {
     var generatedOutputs =
         await _generate(library, generators, buildStep).toList();
 
-    var contentBuffer = new StringBuffer();
+    // Don't output useless files.
+    //
+    // NOTE: It is important to do this check _before_ checking for valid
```
natebosch: Would prefer if this was protected by a test rather than a comment. Can you file an issue so we can followup later?

matanlurey: https://github.com/dart-lang/source_gen/issues/192

kevmoo:
@matanlurey – were you going to publish?

matanlurey:
I'm trying to land .9 with patches still - once I do we can release .10. That way we don't need a .11 if I missed something :)

kevmoo:
You're a wise man...

PR #193, opened 2017-06-28
More json

kevmoo:


lib/generators/json_serializable_generator.dart
```diff
@@ -223,14 +229,58 @@ class JsonSerializableGenerator
...
-      // Then we need to add `?.toList()
-      expression += "?.toList()";
+      if (!isList) {
+        // Then we need to add `?.toList()
+        expression += "?.toList()";
+      }
+
+      return expression;
+    } else if (_coreMapChecker.isAssignableFromType(fieldType)) {
+      var args = _getTypeArguments(fieldType, _coreMapChecker);
```
jakemac53: nit: this method is getting pretty long so it might be worth breaking out each of these sections into their own methods.

lib/generators/json_serializable_generator.dart
```diff
@@ -223,14 +229,58 @@ class JsonSerializableGenerator
...
+          _coreStringChecker.isExactlyType(keyArg);
+
+      var safeValue = false;
+      if (valueType.isDynamic || valueType.isObject) {
+        safeValue = true;
+      } else {
+        safeValue = _isStringBoolNum(valueType);
+      }
+
+      if (safeKey) {
```
jakemac53: there is no else to match here, should it be throwing or is it intended to fall through to the unsafe fallback case?

lib/generators/json_serializable_generator.dart
```diff
@@ -223,14 +229,58 @@ class JsonSerializableGenerator
...
+        if (safeValue) {
+          return expression;
+        }
+
+        // convert
+
```
jakemac53: nit: extra newline, also the comment above is just confusing imo

lib/generators/json_serializable_generator.dart
```diff
@@ -266,22 +316,56 @@ class JsonSerializableGenerator
...
+          _coreStringChecker.isExactlyType(keyArg);
+
+      if (!safeKey) {
+        return "$varExpression /* unsafe key type */";
+      }
+
+      // this is the trivial case. Do a runtime cast to the known type of JSON
+      // map values - `Map<String, dynamic>`
+      if (valueArg.isDynamic || valueArg.isObject) {
+        return "$varExpression as Map<String, dynamic>";
```
jakemac53: keyArg could be `Object` as well though right?

lib/generators/json_serializable_generator.dart
```diff
@@ -266,22 +316,56 @@ class JsonSerializableGenerator
...
+      var output = "($varExpression as List)?.map(($itemVal) => $itemSubVal)";
 
       if (_coreListChecker.isAssignableFromType(searchType)) {
         output += "?.toList()";
       }
 
       return output;
-    }
+    } else if (_coreMapChecker.isAssignableFromType(searchType)) {
+      // Just pass through if
```
jakemac53: it seems like there is a lot of duplication here with the above

natebosch:
Want to shift this over to package:json_serializable 😉 

Happy to start a fresh review for you over there.

PR #195, opened 2017-06-28
increase analyzer min version

camsteffen:
`lib/src/revive.dart` line 26 makes a call to DartObjectImpl#getInvocation() which requires [analyzer 0.29.10](https://github.com/dart-lang/sdk/blob/analyzer-0.29/pkg/analyzer/CHANGELOG.md#02910).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.10+1
+
+* Update minimum `analyzer` package to `0.29.10`.
+
 ## 0.5.10
 
 * Bug fixes:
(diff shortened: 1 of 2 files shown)


camsteffen:
I signed it!

jakemac53:
Looks good, do you mind bumping the version in the pubspec (just a +1 release) and adding a changelog entry?

camsteffen:
right now the version is `0.5.11-dev`. Should I change it to `0.5.10+1`?

kevmoo:
@camsteffen that sounds right – we should make sure the CHANGELOG is updated, too

kevmoo:
Published and tagged. Thanks!

Issue #199, opened 2017-06-29
TypeChecker.exactMatchAnnotationsOf


matanlurey:
/cc @kevmoo for thoughts.

Basically, something like:

```dart
$Foo.annotationsOf(someElement, andAssignable: true);
```

Or we could make this the default behavior, and add `onlyExact: true`. WDUT?

natebosch:
What's the use case for `onlyExact`? Seems to me we'd always be interested in assignable?

matanlurey:
Angular specifically, at least right now, doesn't want to match sub-classes of `@Component`.

natebosch:
Ah ok sounds good.

I think it'd be nicer to have two separate methods rather than a method with a boolean argument.

I'm fine with either `annotationsOf` and `exactMatchAnnotationsOf` or `annotationsOf` and `assignableAnnotationsOf` depending on which we think is more common.

matanlurey:
That works for me! It's a FR.

PR #200, opened 2017-07-05
Add annotationsOfExact.

matanlurey:
Closes https://github.com/dart-lang/source_gen/issues/199.

This is a *breaking change*, and has been reflected on the CHANGELOG+pubspec. In reality it should affect a very small number of users, but I would like to get this in so it's used in Angular ASAP (we support inheriting from `@Component` right by accident, we don't want that).


kevmoo:
No new tests?

matanlurey:
Opened https://github.com/dart-lang/source_gen/issues/204.

Issue #198, opened 2017-06-29
When TypeChecker can't resolve annotation, fail nice


matanlurey:
Today it isn't very friendly:

```
 NoSuchMethodError: The getter 'type' was called on null.
  Receiver: null
  Tried calling: type
  dart:core                                                                    Object.noSuchMethod
  package:source_gen/src/type_checker.dart 55:37                               TypeChecker.annotationsOf.<fn>
  dart:core                                                                    Iterable.isEmpty
  package:source_gen/src/type_checker.dart 49:20                               TypeChecker.firstAnnotationOf
```

PR #201, opened 2017-07-05
Nicer failures when an annotation is not resolvable

matanlurey:
Branched from https://github.com/dart-lang/source_gen/pull/200.

Closes https://github.com/dart-lang/source_gen/issues/198.

### DIFF
--- a/lib/src/type_checker.dart
+++ b/lib/src/type_checker.dart
@@ -38,7 +38,7 @@ abstract class TypeChecker {
   /// package like in the `dart:` SDK.
   const factory TypeChecker.fromUrl(dynamic url) = _UriTypeChecker;
 
-  /// Returns the first constant annotating [element] that is this type.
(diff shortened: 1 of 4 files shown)


lib/src/type_checker.dart
```diff
@@ -49,18 +49,40 @@ abstract class TypeChecker {
     return results.isEmpty ? null : results.first;
   }
 
-  /// Returns every constant annotating [element] that is this type.
+  /// Returns the first constant annotating [element] that is exactly this type.
+  DartObject firstAnnotationOfExact(Element element) {
+    if (element.metadata.isEmpty) {
+      return null;
+    }
+    final results = annotationsOfExact(element);
```
kevmoo: or `.single(orElse: () => null` – or something like that

natebosch: Would be `results.firstWhere((_) => true, orElse: () => null);` IMO what is already in this PR is more readable.

Issue #203, opened 2017-07-06
Add "spanForElement" (for nicer debugging)


matanlurey:
```dart
SourceSpan spanForElement(Element e, SourceFile file) => ...
```

The idea here is to have better error messages that underlying node in a file.

PR #206, opened 2017-07-08
Add `spanForElement`, for nicer debugging messages.

matanlurey:
Closes https://github.com/dart-lang/source_gen/issues/203.

I had to get a little creative for environments where the exact source information is not provided as part of the element structure (i.e. Bazel + Summaries I think), so I don't mind annotating this `@experimental` and doing more validation before we settle on this implementation.

/cc @natebosch too, since I think @jakemac53 is out for a bit.

### DIFF
--- a/lib/src/span_for_element.dart
+++ b/lib/src/span_for_element.dart
@@ -1,0 +1,41 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/element/element.dart';
+import 'package:source_span/source_span.dart';
+
+import 'utils.dart';
+
+/// Returns a source span that spans the location where [element] is defined.
+///
+/// May be used to emit user-friendly warning and error messages:
+/// ```dart
+/// void invalidClass(ClassElement class) {
+///   log.warning(spanForElement.message('Cannot implement "Secret"'));
+/// }
+/// ```
+///
+/// Not all results from the analyzer API may return source information as part
+/// of the element, so [file] may need to be manually provided in those cases.
+SourceSpan spanForElement(Element element, [SourceFile file]) {
+  final url = assetToPackageUrl(element.source.uri);
+  if (file == null) {
+    final contents = element?.source?.contents;
+    if (contents == null) {
+      return new SourceSpan(
+        new SourceLocation(
+          element.nameOffset,
+          sourceUrl: url,
+        ),
+        new SourceLocation(
+          element.nameOffset + element.nameLength,
+          sourceUrl: url,
+        ),
+        element.name,
+      );
+    }
+    file = new SourceFile.fromString(contents.data, url: url);
+  }
+  return file.span(element.nameOffset, element.nameOffset + element.nameLength);
+}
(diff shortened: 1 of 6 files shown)


lib/src/span_for_element.dart
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/element/element.dart';
+import 'package:source_span/source_span.dart';
```
kevmoo: Missing explicit dependency on `source_span` here...

lib/src/utils.dart
```diff
@@ -122,6 +122,20 @@ Uri packageToAssetUrl(Uri url) => url.scheme == 'package'
           ..addAll(url.pathSegments.skip(1)))
     : url;
 
+/// Returns a `asset:` URL converted to a `package:` URL.
```
kevmoo: Where is the `asset:` uri logic defined? For barback?

Might be nice to clarify this...

test/span_for_element_test.dart
```diff
@@ -0,0 +1,31 @@
...
+      library test_lib;
+      
+      abstract class Example implements List {}
+    ''', inputId: new AssetId('test_lib', 'lib/test_lib.dart'));
+    library = resolver.getLibraryByName('test_lib');
+  });
+
+  test('should highlight the use of "class Example"', () {
+    expect(
+        spanForElement(library.getType('Example')).message('Here it is'),
```
kevmoo: That's pretty cool!

lib/src/span_for_element.dart
kevmoo: Well, not here – in pubspec

matanlurey:
> Sure...I guess...
> 
> I'm getting a bit...uh...nervous/concerned that `source_gen` is evolving into a chimera between an opinionated model for pkg/build and a set of analyzer helpers...

@kevmoo I hear you. At the same time it makes the churn around creating code generators much less (less points of failure for the analyzer team and Nate/Jake). I think once we are in a "2.0 world" it might make sense to branch into a pkg/analysis. 

We could do it now too but I don't know if there is a huge value. WDUT?

kevmoo:
Eh...we can hold off. Not a big deal. Need to pull out the JSON bits, too.

lib/src/span_for_element.dart
matanlurey: Your wish is my command,

lib/src/utils.dart
matanlurey: Your wish is my command,

lib/src/utils.dart
```diff
@@ -127,6 +127,10 @@ Uri packageToAssetUrl(Uri url) => url.scheme == 'package'
 /// For example, this transformers `asset:source_gen/lib/source_gen.dart' into:
 /// `package:source_gen/source_gen.dart`. Asset URLs that aren't pointing to a
 /// file in the 'lib' folder are not modified.
+///
+/// Asset URLs come from `package:build`, as they are able to describe URLs that
+/// are not describable using `package:...`, such as files in the `bin`, `tool`,
+/// `web`, or even root directory of a package - `asset:some_lib/web/main.dart`.
```
kevmoo: `asset:some_lib` -> `asset:some_pkg`, right?

matanlurey:
Oops I can make that change via GitHub (or you can) sorry!

PR #218, opened 2017-07-19
Split out GeneratorBuilder interface

natebosch:
Add a PartBuilder to handle multiple generators and part files, and
SourceBuilder to handle single generator standalone files.

Replaces the `isStandalone` argument with an explicit interface and
removes the hidden dependency on the number of generators.

### DIFF
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -15,16 +15,7 @@ import 'utils.dart';
 /// Returns [generatedCode] formatted, usually with something like `dartfmt`.
 typedef String OutputFormatter(String generatedCode);
 
-/// Wraps multiple [Generator]s and exposes them as a [Builder] for tooling.
-///
-/// ```dart
-/// // Creates a new builder that the following two generators.
-/// new GeneratorBuilder([
-///   new JsonSerializableGenerator(),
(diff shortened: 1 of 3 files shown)


natebosch:
Sending out for some initial feedback - I've got other breaking changes I'd like to make before this gets published.

lib/src/builder.dart
```diff
@@ -189,6 +136,80 @@ Stream<GeneratedOutput> _processUnitMember(
...
+    }
+    contentBuffer.writeln('part of $name;');
+    contentBuffer.writeln();
+  }
+
+  @override
+  String toString() => 'PartBuilder:$generators';
+}
+
+/// Wraps a [Generator] and exposes it as a [Builder] which produces a Dart
```
matanlurey: Does it make sense to just have this be `Builder` instead of something explicitly different?

natebosch: How do you mean?

I also considered keeping a single class with two constructors, `GeneratorBuilder` and `GeneratorBuilder.standalone`. Is that what you're suggesting?

natebosch:
We're going to end up looking a little different after #223 and #220 - closing for now.

PR #225, opened 2017-07-19
Delete JsonSerializable and related code

### LINKED ISSUE #222: Delete the JSON generator stuff
It doesn't really fit with the package. Blocked on getting this published in some other package.

natebosch:
Closes #222 

This has all been moved to package:json_serializable

- Drop the annotations and generators
- Drop the tests
- Add _test_annotations for one use case which was looking at a removed
  annotation


natebosch:
We'll just need to make sure we do a pub publish of package:json_serializable before publishing this.

natebosch:
test_all.dart was probably used for coverage - we aren't depending on that anymore are we?

natebosch:
Hmm, `ensure_build_test.dart` wants me to not have moved away the `tool/build.dart`.

I might need to add back some kind of E2E test here...

natebosch:
Actually, this test doesn't seem powerful enough to justify the extra overhead. I think I'll drop it.

PR #226, opened 2017-07-19
Drop the safeIterate helper

natebosch:
This existed so that we could separate the exceptions that occur during
iteration from the exceptions that occur during processing since a for
loop can't have a try/catch around only getting elemetns. This is more
easily expressed with a try/catch around `Iterable.toList()`.

### DIFF
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -162,10 +162,14 @@ class GeneratorBuilder extends Builder {
 
 Stream<GeneratedOutput> _generate(LibraryElement unit,
     List<Generator> generators, BuildStep buildStep) async* {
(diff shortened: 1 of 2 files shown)


natebosch:
This does change behavior slightly - we no longer generate *any* code if resolution fails for any element. This difference is just academic - if resolution fails (and therefore the build fails) we don't care about any of the output.

kevmoo:
Are you sure the behavior you were going for here – https://github.com/dart-lang/source_gen/commit/d3201fb28026773ea35f97b18386ad57d41a56c7 – is preserved?

natebosch:
Yup, behavior is fine since we don't have the try/catch around the processing step.

PR #231, opened 2017-07-20
Change Generator to run on LibraryElement

### LINKED ISSUE #220: Add `GeneratorForLibrary`
The Angular generator only operates on `LibraryElement` anyway. Most likely the only way a 'standalone' generator makes sense is if it's a single `GeneratorForLibrary` rather than operating on multiple elements.

We probably want Part builders to be able to take one or more `GeneratorForLibrary` _or_ one or `Generator`.

### LINKED ISSUE #223: Find a solution for single-add situations like utility methods

### LINKED ISSUE #224: Remove ability to instantiate annotations
Instead of passing a `T` we could pass a `DartObject` which we guarantee (using the type matcher) will represent a `T`.
The API for using it is not as nice (`.getField('name')` instead of `.name`) but it would mean we don't have a contract we can't always satisfy and it makes the code simpler.

I'm not completely convinced we should do this, but it's worth discussing. There has been talk about getting more reliable instantiation from `DartObject` in the analyzer itself - but this hasn't gone anywhere and is unlikely to for a while, so I'm not sure we should bank on it.

cc @kevmoo @jakemac53

natebosch:
Closes #224, closes #220, closes #223

This is a breaking behavior change that may be slightly harder to notice
than usual, because the old method signature is still a valid override.

- Delete annotations.dart and tests. Annotations are no longer
  instantiated using mirrors, the DartObject is used isntead.
- Move iterating over Elements into GeneratorForAnnotation instead of
  builder.dart
- Drop `sourceMember` from `GeneratedOutput` since an entire output
  relates to a library.
- Update test Generator implementations to operate on a library and
  iterate over elements manually when necessary.

Other cleanup:
- Remove some unnecessary library statements


natebosch:
😢 I broke const-ness with this - doesn't work with package:json_serializable as is

natebosch:
https://github.com/dart-lang/json_serializable/pull/17 shows what this would look like in package:json_serializable

lib/src/generator.dart
```diff
@@ -2,30 +2,21 @@
...
-///
-/// Clients should _extend_ this class and _override_ [generate]:
-/// ```dart
-/// class MyGenerator extends Generator {
-///   Future<String> generate(Element element, BuildStep buildStep) async {
-///     // Return a string representing the code to emit.
-///   }
-/// }
-/// ```
 abstract class Generator {
```
kevmoo: You know, some docs here might be nice...

natebosch: For the most part the doc on `generate` is the important part...

natebosch: Added a short doc that points toward `[generate]`.

Issue #219, opened 2017-07-19
Make all members of `GeneratorBuilder` (or whatever it ends up getting called) private


natebosch:
Currently things like `generatedExtension` and `generators` are public, but we should probably keep the interface to that of `Builder` unless we find a use case for reading something else externally. We don't want to handcuff ourselves to more breaking changes if we tweak the implementation.

natebosch:
Also probably make the typedef for the formatter private.

kevmoo:
...new function syntax FTW!

natebosch:
When can we start using new function syntax? Is it fully available in `1.24`? Do we feel OK setting our SDK constraint that high?

matanlurey:
Fully available in 1.24. Angular has that restriction FWIW.

PR #234, opened 2017-07-21
Add separate classes for Part vs Library builders

### LINKED ISSUE #117: Make StandaloneBuilder a separate class from GeneratorBuilder

natebosch:
Closes #117

Fixes the API weirdness where `isStandalone` only works with a single
Generator, and shouldn't be passed with `requireLibraryDirective`.

This is a small step - we still need to clean up the interface (#219).

- Make `GeneratorBuilder` private as `_Builder`.
- Subclass as `PartBuilder` and `LibraryBuilder` with only forwarding
  constructors to give nicer signatures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,7 @@
   * `GeneratorForAnnotation` passes in a `ConstantReader` for the annotation
     instance rather than re-creating it using mirrors.
   * `GeneratorBuilder` is replaced with `PartBuilder` and `LibraryBuilder`
-    depending on whether the `Generator`(s) used are meant to be included in a
-    `part` file.
+    depending on whether the output is meant to be included in a `part` file.
   * Removed `JsonSerializable` and related classes. These are moved to
     `package:json_serializable`.
   * Removed `lib/builder.dart`. Import through `source_gen.dart` instead.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
     continue to call `generateForAnnotatedElement` repeatedly for each element.
   * `GeneratorForAnnotation` passes in a `ConstantReader` for the annotation
     instance rather than re-creating it using mirrors.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
     continue to call `generateForAnnotatedElement` repeatedly for each element.
   * `GeneratorForAnnotation` passes in a `ConstantReader` for the annotation
     instance rather than re-creating it using mirrors.
+  * `GeneratorBuilder` is replaced with `PartBuilder` and `LibraryBuilder`
+    depending on whether the `Generator`(s) used are meant to be included in a
```
kevmoo: Tough to parse this.

`depending on...` -> `depending on whether the target is a part file or a stand-alone library.` Maybe?

test/builder_test.dart
```diff
@@ -32,46 +32,25 @@ void main() {
...
-        new GeneratorBuilder([const CommentGenerator()], isStandalone: true);
+    var builder = new LibraryBuilder(const CommentGenerator());
     await testBuilder(builder, srcs,
         generateFor: new Set.from(['$pkgName|lib/test_lib.dart']),
         outputs: {
           '$pkgName|lib/test_lib.g.dart': _testGenStandaloneContent,
         });
   });
 
-  test('Generate explicitly non-standalone output file', () async {
```
kevmoo: Did you mean to delete these entirely?

natebosch: yeah, they are testing conditions which it is no longer possible to hit

CHANGELOG.md
natebosch: Reworded

kevmoo:
LGTM

PR #247, opened 2017-07-25
Don't emit library in allElements

### LINKED ISSUE #245: Don't output the `LibraryElement` from libraryReader.allElements
I think at this point it will most likely just be skipped since those who want the library will use `LibraryReader.element` and most generators are unlikely to treat the `LibraryElement` like any other `element`...

natebosch:
Fixes #245

The library is available through `element` and most Generators will want
to make a distinction anyway.

### DIFF
--- a/lib/src/library.dart
+++ b/lib/src/library.dart
@@ -29,7 +29,6 @@ class LibraryReader {
   /// All of the declarations in this library, including the [LibraryElement] as
   /// the first item.
   Iterable<Element> get allElements sync* {
-    yield element;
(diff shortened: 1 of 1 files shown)


lib/src/library.dart
```diff
@@ -29,7 +29,6 @@ class LibraryReader {
   /// All of the declarations in this library, including the [LibraryElement] as
```
jakemac53: Looks like the comment needs to be updated as well

natebosch: Good catch. Done.

Issue #248, opened 2017-07-25
Add optional arg to skip unresolved annotations for TypeChecker methods


jakemac53:
Specifically, the following apis will get a `bool skipUnresolvedAnnotations` optional parameter:

`TypeChecker#firstAnnotationOf`,
`TypeChecker#firstAnnotationOfExact`
`TypeChecker#annotationsOf`
`TypeChecker#annotationsOfExact`

This allows you to use the `TypeChecker` in a partially resolved world to match annotations for which you do have resolved type information.

natebosch:
Let's call it `skipUnresolved` since `Annotation` is already known by context.

matanlurey:
👍

jakemac53:
Actually, how about `throwOnUnresolved`? It is a bit more explicit about the failure case I think.

PR #249, opened 2017-07-25
add `throwOnUnresolved` to TypeChecker methods to allow matching annotations in a non-fully resolved context

jakemac53:
Closes https://github.com/dart-lang/source_gen/issues/248


jakemac53:
updated

PR #266, opened 2017-07-27
Move utils for constants...

matanlurey:
... and add some tests, so we can catch any analyzer changes in Dart 1 -> 2.


natebosch:
Just to make sure I'm clear - this only moves code around and adds tests, this doesn't change the public surface area of the package right? In that case 👍

matanlurey:
No public interface changes, yup!

PR #271, opened 2017-08-16
Update to build_test v0.7

natebosch:


test/type_checker_test.dart
```diff
@@ -31,33 +31,35 @@ void main() {
...
+    LibraryElement core;
+    LibraryElement collection;
+    LibraryReader sourceGen;
+    await resolveSource(r'''
       export 'package:source_gen/source_gen.dart';
-    ''', tearDown: resolverDone.future);
+    ''', (resolver) async {
+      core = await resolver.findLibraryByName('dart.core');
+      collection = await resolver.findLibraryByName('dart.collection');
+      sourceGen =
```
natebosch: This is a case where we could still use `tearDown` if we want to keep the variables `final` - I'm open to either approach.

jakemac53: This is fine imo, these are local variables so `final` is more personal style than anything anyways.

PR #276, opened 2017-10-08
new feature, linted, spell-checked

kevmoo:


lib/src/builder.dart
```diff
@@ -123,6 +127,10 @@ class _Builder extends Builder {
 
     var genPartContent = contentBuffer.toString();
 
+    if (_header.isNotEmpty) {
+      genPartContent = '${_header}\n$genPartContent';
```
jakemac53: nit: move this up above and add it to `contentBuffer` right after creation, should avoid some extra string copying and be a bit cleaner.

lib/src/constants/reader.dart
```diff
@@ -122,7 +123,7 @@ abstract class ConstantReader {
 
 class _NullConstant extends ConstantReader {
   @alwaysThrows
-  static dynamic _throw(String expected) {
+  static T _throw<T>(String expected) {
```
jakemac53: why not just set `void` as the return type?

lib/src/utils.dart
```diff
@@ -69,7 +69,7 @@ String suggestLibraryName(AssetId source) {
   parts[parts.length - 1] = parts.last.split('.').first;
   // [lib/test] --> [test]
   if (parts.first == 'lib') {
-    parts = parts.skip(1);
+    parts.removeAt(0);
```
jakemac53: nit: the skip is probably actually faster in this scenario

lib/src/constants/reader.dart
kevmoo: Because then it makes strong mode sad. The return value is needed since this is used in ` => ` calls

lib/src/utils.dart
kevmoo: Yes, but then we're trying Iterables like Lists...

lib/src/builder.dart
kevmoo: kk

lib/src/constants/reader.dart
jakemac53: ah I see ok

lib/src/utils.dart
jakemac53: ah ya ok probably not worth going through the type shenanigans just for a minor optimization here

kevmoo: t'was my thought

Issue #292, opened 2018-01-12
Revert prefixFor feature(s).


matanlurey:
This is code that is super tough to get right, and I imagine most of the time most generators that use part files just disallow using prefixes to import code because of it. In AngularDart we still need to support this right now, and our code is brittle.

Adding this feature would also make analyzer changes more visible.

```dart
class LibraryReader {
  /// Returns the identifier prefix if [element] is referenced by one.
  ///
  /// For example in the following file:
  /// ```dart
  /// import 'bar.dart' as bar;
  ///
  /// const b = const bar.Bar();
  /// ```
  ///
  /// If there is no prefix, one could not be computed, or the [element] is
  /// in a library different from the current one this will return `null`.
  String prefixFor(Element element);
}
```

`prefixFor` could return `bar` for the top-level field `b`.

Other considerations:
* [ ] If the element is outside the current library, always return `null`. This consideration is important because in implementations such as `bazel_codegen`, the underlying function we will need to use, `computeNode()`, fails on files provided only by summaries.

/cc @natebosch to spot check this feature's acceptability :)
/cc @MichaelRFairhurst who helped me identify why this fails in Angular (summary file)

natebosch:
Sounds like a good feature.

> `prefixFor` could return `bar` for the top-level field `b`.

Hmm, the field doesn't have a prefix does it? It makes sense to me that the `ClassElement` for `Bar` would return a prefix of `'bar'` - but I don't understand why `b` would be considered to have a prefix.

> If the element is outside the current library, always return `null`

I think we should throw instead since `null` means 'no prefix'

matanlurey:
I've decided to revert this feature, and remove it.

It's not trivial to support in future source-generation models.

PR #296, opened 2018-01-12
Start adding LibraryReader.prefixForType.

matanlurey:
Work towards https://github.com/dart-lang/source_gen/issues/292.

I've made the feature package-private so we can iterate. It's possible we won't even be able to launch this as-is (if `computeNode()` explodes), but at least we will have the test case(s) when we have a replacement API, so I think it's worth checking (most) of this in.

### DIFF
--- a/lib/src/library.dart
+++ b/lib/src/library.dart
@@ -74,18 +74,60 @@ class LibraryReader {
   Iterable<ClassElement> get classElements =>
       element.definingCompilationUnit.types;
 
-  Iterable<Element> _getElements(CompilationUnitMember member) {
+  static Iterable<Element> _getElements(CompilationUnitMember member) {
     if (member is TopLevelVariableDeclaration) {
       return member.variables.variables
           .map(resolutionMap.elementDeclaredByVariableDeclaration);
     }
     var element = resolutionMap.elementDeclaredByDeclaration(member);
-
     if (element == null) {
-      print([member, member.runtimeType, member.element]);
-      throw new Exception('Could not find any elements for the provided unit.');
+      throw new StateError(
+          'Could not find any elements for the provided unit.');
     }
-
     return [element];
   }
+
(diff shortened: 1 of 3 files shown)


lib/src/library.dart
```diff
@@ -74,18 +74,60 @@ class LibraryReader {
...
+  /// Returns the identifier prefix of [element] if it is referenced by one.
+  ///
+  /// For example in the following file:
+  /// ```
+  /// import 'bar.dart' as bar;
+  ///
+  /// const b = const bar.Bar();
+  /// ```
+  ///
+  /// ... we'd assume that `b`'s element has a prefix of `bar`.
```
natebosch: I'm still confused by this. There isn't anywhere we'd have the code `bar.b` - so why is it that `b` is considered to have a prefix?

test/library_reader_test.dart
```diff
@@ -0,0 +1,73 @@
...
+
+  group('top level fields', () {
+    List<Element> topLevelFieldTypes;
+
+    setUpAll(() {
+      topLevelFieldTypes = b.element.definingCompilationUnit.topLevelVariables;
+    });
+
+    test('should read the prefix of a type', () {
+      expect(testingPrefixForType(b, topLevelFieldTypes[0]), 'a_prefixed');
```
natebosch: Can we look for specific variables rather than rely on the ordering?

lib/src/library.dart
matanlurey: Yeah this isn't an accurate example anymore. I'll update it.

test/library_reader_test.dart
matanlurey: Sure.

matanlurey: Done.

lib/src/library.dart
matanlurey: Done.

Issue #293, opened 2018-01-12
LibraryReader: Add .pathTo(<AssetId|Element|Uri>)


matanlurey:
Like https://github.com/dart-lang/source_gen/issues/292, this is code that is non-trivial to get right, and even more difficult when you consider our current code generation strategies, which rely on `asset:` URLs to identify files in non-lib directories of packages.

```dart
class LibraryReader {
  /// Basically the same thing, as a convenience to use with assets.
  Uri pathToAsset(AssetId to);

  /// Basically the same thing, as a convenience to use with elements.
  Uri pathToElement(Element to);

  /// Returns a [Uri] from the current library to the one described in [to].
  ///
  /// If possible, `package:` or `dart:` annotation will be used to reference
  /// the file, falling back to relative paths if required (such as in the `test`
  /// directory).
  ///
  /// May throw [ArgumentError] if it is not possible to resolve a path.
  Uri pathToUrl(Url to);
}
```

/cc @natebosch again, to spot-check this feature.

natebosch:
Makes sense to me. I'm guessing this will handle any windows weirdness?

matanlurey:
That's the hope!

PR #297, opened 2018-01-12
Add LibraryReader.pathToUrl.

matanlurey:
Partial work towards https://github.com/dart-lang/source_gen/issues/293.

I omitted `pathToAsset` and `pathToElement`, because (a) they are just conveniences and (b) this is the bulk of the code for first review - the latter should be quite quick.

### DIFF
--- a/lib/src/library.dart
+++ b/lib/src/library.dart
@@ -78,6 +78,11 @@ class LibraryReader {
   /// the library, falling back to relative paths if required (such as in the
   /// `test` directory).
   ///
+  /// The support [Uri.scheme]s are (others throw [ArgumentError]):
+  /// * `dart`
+  /// * `package`
+  /// * `asset`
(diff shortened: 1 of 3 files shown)


lib/src/library.dart
```diff
@@ -70,6 +72,77 @@ class LibraryReader {
     }
   }
 
+  /// Returns a [Uri] from the current library to the one provided.
+  ///
+  /// If possible, a `package:` or `dart:` URL scheme will be used to reference
+  /// the library, falling back to relative paths if required (such as in the
+  /// `test` directory).
+  ///
+  /// May throw [ArgumentError] if it is not possible to resolve a path.
```
natebosch: This should mention the schemes it supports

matanlurey: Done! Thanks.

PR #312, opened 2018-02-15
Add toString overrides for Builders

### LINKED ISSUE #311: Implement toString() in LibraryBuilder,PartBuilder.
_From @matanlurey on February 15, 2018 17:37_

See errors like https://github.com/dart-lang/angular/issues/897:

> [WARNING] Instance of 'LibraryBuilder' on ...

There is nothing here that says it is _Angular_ that is warning, versus say, DDC.

I could override `toString()`, but that also means `LibraryBuilder`/`PartBuilder` in `package:source_gen` would need to delegate `toString()` to the wrapper generator; maybe that is OK too?

_Copied from original issue: dart-lang/build#1023_

natebosch:
Closes #311

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.7.6
+
+* Include Generator types in `toString()` on builder implementations. This will
+  improve the error messages in `build_runner`.
+
 ## 0.7.5
 
 * The `PartBuilder` constructor parameter `requireLibraryDirective` now defaults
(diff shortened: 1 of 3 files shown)


natebosch:
Note that `Generator` already uses runtimeType in it's toString so for most users this will already point in a good direction.

https://github.com/dart-lang/source_gen/blob/a2ef2c33fe64c29d1ed60e8e419b9098cbce10bc/lib/src/generator.dart#L26

Generator implementations can also override this if they want something displayed other than their type name.

PR #323, opened 2018-04-09
Code cleanup

kevmoo:
Preparing for new features

### DIFF
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -94,10 +94,7 @@ class _Builder extends Builder {
 
     if (!_isStandalone) {
       var asset = buildStep.inputId;
-      var name = nameOfPartial(
-        library,
-        asset,
-      );
+      var name = nameOfPartial(library, asset);
       if (name == null) {
         var suggest = suggestLibraryName(asset);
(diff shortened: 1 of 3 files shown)


lib/src/builder_impl.dart
```diff
@@ -0,0 +1,173 @@
...
+import 'generator.dart';
+import 'library.dart';
+import 'utils.dart';
+
+const defaultFileHeader = '''// GENERATED CODE - DO NOT MODIFY BY HAND
+
+''';
+
+/// A [Builder] wrapping on one or more [Generator]s.
+class BuilderImpl extends Builder {
```
natebosch: why does this need to be in a different file?

why does this need to be named like "Impl"?

kevmoo:
Because that's the way *uh huh uh huh* I like it. *uh huh uh huh*

lib/src/builder_impl.dart
matanlurey: +1. This change isn't valuable.

kevmoo:
Ack all around – @natebosch @matanlurey – PTAL

PR #324, opened 2018-04-09
added optional `associatedElement` param to InvalidGenerationSourceError

kevmoo:


lib/src/generator.dart
```diff
@@ -36,9 +38,26 @@ class InvalidGenerationSourceError extends Error {
   /// May be an empty string if unknown.
   final String todo;
 
-  InvalidGenerationSourceError(this.message, {String todo})
+  /// The code element associated with this error.
+  ///
+  /// May be `null` if the error had no associated element.
+  final Element associatedElement;
```
natebosch: I'm not sure if the word "associated" is making this more clear than it would be with only "element"

kevmoo:
ack – PTAL

PR #326, opened 2018-04-13
Add sortedFields and isEnum utils

kevmoo:
And remove now unneeded (and not public anyway) `friendlyNameForElement`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 * `InvalidGenerationSourceError` added an optional `element`
   parameter to support more helpful error messages.
 
+* Added a new top-level function `sortedFields`. Useful when generating code
+  that wants to align with the field order in a `ClassElement`.
+
 ## 0.8.0
 
 * **BREAKING** removed the deprecated `requireLibraryDirective` parameter in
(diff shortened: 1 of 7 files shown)


kevmoo:
I've needed `sortedFields` twice

https://github.com/kevmoo/build_cli/blob/master/build_cli/lib/src/to_share.dart#L144
https://github.com/dart-lang/json_serializable/blob/master/json_serializable/lib/src/utils.dart#L144

So there is a need.

I'm trying to centralize this non-trivial logic to help others – and hopefully solve it one place.

matanlurey:
I understand. I still don't think we should use any more private implementation details given that the analyzer team is going to be (and is) working very hard to make a new stable API on top of the CFE.

natebosch:
`s/needed/wanted`

I understand that it's _nice_ to have generated code use the same ordering as the source - what would be _broken_ without it?

matanlurey:
We've already put our feet down and blocked many changes here and in angular and elsewhere.

Neither of those two packages are P1s for our customers, but keeping `source_gen` running is. Feel free to schedule a VC if you'd like to discuss these concerns more in person.

PR #330, opened 2018-05-07
Add dartStringLiteral helper

kevmoo:
Replaces code duplicated in `json_serializable` and `build_cli`.  Can also replace `escapeSingleQuoteString` in angular `abstract_emitter.dart`.


kevmoo:
> I'd want this functionality built in to literalString if possible...

I think I agree. The way `literalString` is setup, though, makes it weird. I don't think the current `raw` parameter makes sense.

This is meant to at least cover json_serializable, angular (and build_cli)

We'd need to chat w/ @matanlurey about how/where to put it in code_builder.

At the moment json_serializable doesn't use code_builder – but in theory it could/should...

kevmoo:
Will look into the tiny package...

PR #335, opened 2018-06-09
Export log from pkg:build

kevmoo:
Also removed unused dev dependency on cli_util

Will publish once approved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.8.2
+
+* Expose the top-level `log` property exported from `package:build`. 
+
 ## 0.8.1+3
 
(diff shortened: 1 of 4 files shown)


natebosch:
I don't think any consumers of source_gen will get by without importing build... How does this help us?

kevmoo:
I find myself constantly importing `build` just to get `log` – and deleting the import when I'm done debugging.

I also find myself importing `source_gen` in for more places than build when I'm creating a Generator.

...so I thought it would be useful

kevmoo:
> How do you avoid the import for BuildStep?

I don't. I just find (for json_serializable) I import source_gen in many more places.

I can give up on this – not a biggy.

PR #337, opened 2018-06-10
cleanup: move to mono_repo

kevmoo:
Will land as a single commit once https://github.com/dart-lang/source_gen/pull/336 lands


natebosch:
What is the root level `README` link accomplishing for us here?

kevmoo:
keeping us from having to duplicate content? Or keep something up-to-date. 🤷

natebosch:
Oh interesting, this does make the landing page on github nicer. It doesn't work for the `/blob` URLs which is what I had been looking at.

natebosch:
I think it does probably make sense for us to link the README by default when there is a clear "main" package. I'll do that in `test`. https://github.com/dart-lang/test/pull/1477

Issue #319, opened 2018-03-21
Add a CombiningBuilder for .g.dart files and document how to use it


natebosch:
We added `runs_before` and `applies_builder` to build config so it should be possible now to put everything together with one or more source_gen part builders cooperating to generate a single part file. We'll want a "CombiningBuilder" ("MergingBuilder"?) in this package to read in all the files generated by different part builders and write them to a single output.

- [x] Fix duplicate headers - https://github.com/dart-lang/source_gen/pull/348
- [x] Fix logging of missing `part` directive - https://github.com/dart-lang/source_gen/pull/348
- [x] Document usage in readme
- [x] Update `build.yaml` to properly expose `CombiningBuilder`
- [x] **BLOCKING** Get build issue fixed related to running built-to-cache builder over transitive dependencies - https://github.com/dart-lang/build/issues/1578
- [x] Add a cleanup builder

After published...
- [x] Add a working example to `repo/example`

kevmoo:
@grouma – added some work items before we can close this. The first two are handled by my pending PR

kevmoo:
How are we doing w/ the check boxes here @natebosch and @grouma ?

kevmoo:
...still need to tweak build.yaml, right?

natebosch:
> ...still need to tweak build.yaml, right?

I think it looks correct (other than a nit about single vs double quotes). README and example are the remaining issues. I wanted also to validate usage which is what https://github.com/dart-lang/json_serializable/pull/260 is meant to accomplish.

kevmoo:
ack

natebosch:
yes, I think we are safe to publish now. @grouma, @jakemac53 - can you think of anything else blocking?

jakemac53:
I haven't been following this to closely :grimacing: but nothing blocking I am aware of.

kevmoo:
Version bumb to v1 first?

grouma:
Should be good to go.

natebosch:
There's an example now, I think we've finished.

PR #341, opened 2018-06-12
feature: allow Generators to return multiple values

kevmoo:
Supported return values include a single `String` or multiple Strings
within an `Iterable` or `Stream`. It is also valid to return a `Future`
of any of the above.

Duplicate values are also collapsed across Generators within a Builder.
This allows creating shared helpers.

Progress towards https://github.com/dart-lang/source_gen/issues/319


source_gen/lib/src/generator.dart
```diff
@@ -18,11 +18,15 @@ abstract class Generator {
 
   /// Generates Dart code for an input Dart library.
   ///
-  /// May create additional outputs through the `buildStep`, but the 'primary'
-  /// output is Dart code returned through the Future. If there is nothing to
-  /// generate for this library may return null, or a Future that resolves to
-  /// null or the empty string.
-  FutureOr<String> generate(LibraryReader library, BuildStep buildStep) => null;
+  /// Supported return values include a single [String] or multiple [String]
```
natebosch: I'm not sure the reliance on Doc comments over type signatures is worth it in this case. How does this relate to the goal of a CombiningBuilder? I don't think that implementation needs to use the Generator infrastructure at all.

kevmoo: It turns out to be *super* useful in the single case.

For instance, you can now emit helpers for each element in a generator and they get collapsed into one. So you don't need to (1) put them inline in a function or (2) put them in the "annotations" package.

I have a design of the combining builder as just a natural extension of the existing Part/Library builder impl, but I wanted to get this in first.

natebosch: @grouma was going to take a look at the CombiningBuilder

>For instance, you can now emit helpers for each element in a generator and they get collapsed into one. So you don't need to (1) put them inline in a function or (2) put them in the "annotations" package.

I'm not entirely convinced that emitting helpers this way is the most understandable way to handle it. It might work smoothly for us when we have a deep understanding of this package but  expands the surface area in a way that might be hard to document for other users.

> I have a design of the combining builder as just a natural extension of the existing Part/Library builder impl, but I wanted to get this in first.

The CombiningBuilder shouldn't need a `LibraryReader` (though the overall builder will probably want to check for the `part` directive and will want to use a Resolver) and I don't quite understand how it fits the `Generator` interface.

natebosch: Let's maybe chat about this when you're back in the office, or @grouma could review this if he thinks it will be useful.

grouma: I don't have the context yet to determine if this will be useful. Let's hold off on submitting it. I'll be diving into https://github.com/dart-lang/source_gen/issues/319 today.

kevmoo:
@natebosch – scoped way down to just `GenerateForAnnotation` – which is still super useful

source_gen/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@
   not be `.g.dart`. To produce `.g.dart` files please use the
   `SharedPartBuilder`.
 
+* `GeneratorForAnnotation.generateForAnnotatedElement` now allow multiple return
```
jakemac53: allows*

source_gen/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@
...
+  values when implementations return an `Iterable` or `Stream`.
+  * Values from `generateForAnnotatedElement` have whitespace trimmed. `null`
```
jakemac53: its it worth allowing nulls? that seems like its just hiding errors

kevmoo:
https://travis-ci.org/dart-lang/source_gen/jobs/392434788#L531 – weird!

@jakemac53 – do you have -dev.62 on your machine? Could you try this?

source_gen/lib/src/generator_for_annotation.dart
```diff
@@ -34,18 +35,36 @@ abstract class GeneratorForAnnotation<T> extends Generator {
...
 
   @override
-  Future<String> generate(LibraryReader library, BuildStep buildStep) async {
-    var elements = library.annotatedWith(typeChecker);
-    var allOutput = await Future.wait(elements.map((e) async =>
-        await generateForAnnotatedElement(e.element, e.annotation, buildStep)));
-    // TODO interleave comments indicating which element produced the output?
-    return allOutput.where((o) => o != null && o.isNotEmpty).join('\n');
+  FutureOr<String> generate(LibraryReader library, BuildStep buildStep) async {
+    var set = await new Stream.fromIterable(library.annotatedWith(typeChecker))
```
jakemac53: Choose better names for these (you are shadowing it below, and it's generally just a bad name lol).

Also isn't this a reserved word?

source_gen/lib/src/output_helpers.dart
```diff
@@ -0,0 +1,64 @@
...
+import 'package:async/async.dart';
+
+/// Runs [run] and converts [Future], [Iterable], and [Stream] implementations
+/// containing [String] to a single [Stream] while ensuring all thrown
+/// exceptions are forwarded through the return value.
+Stream<String> normalizeGeneratorOutput(Object run()) => StreamCompleter
+    .fromFuture(new Future.sync(run).then(_normalizeGeneratorOutput));
+
+Stream<String> _normalizeGeneratorOutput(Object value) {
+  if (value is String) {
```
jakemac53: would this return true for null and create `[null]`? would it matter?

source_gen/lib/src/output_helpers.dart
```diff
@@ -0,0 +1,64 @@
...
+    value = [value];
+  }
+
+  value ??= [];
+
+  if (value is Iterable) {
+    value = safeIteratorToStream(value as Iterable);
+  }
+
+  if (value is Stream) {
```
jakemac53: nit: check for is! stream so the error is closer to the code that is really checking for it?

source_gen/lib/src/output_helpers.dart
kevmoo: `null` is not a String. You end up with an empty array below.

source_gen/lib/src/output_helpers.dart
kevmoo: But then I have to cast the value I just checked – because we don't do inference for values after `is! Stream`.

Okay if I leave it?

source_gen/lib/src/generator_for_annotation.dart
kevmoo: fixed

jakemac53:
> @jakemac53 – do you have -dev.62 on your machine? Could you try this?

tests pass for me locally....

source_gen/CHANGELOG.md
jakemac53: oh so maybe allow a top level null but not individual `null` values in the Stream/Iterable then?

kevmoo: Hrm...

Well, I kinda like `null`

I could have this in my generator:

```dart
yield _myHelper(context);
```

Then `_myHelper` is designed to return null on no-op.

source_gen/lib/src/output_helpers.dart
```diff
@@ -0,0 +1,64 @@
...
+
+import 'package:async/async.dart';
+
+/// Runs [run] and converts [Future], [Iterable], and [Stream] implementations
+/// containing [String] to a single [Stream] while ensuring all thrown
+/// exceptions are forwarded through the return value.
+Stream<String> normalizeGeneratorOutput(Object run()) => StreamCompleter
+    .fromFuture(new Future.sync(run).then(_normalizeGeneratorOutput));
+
+Stream<String> _normalizeGeneratorOutput(Object value) {
```
jakemac53: nit: `dynamic` I think it techinically the right thing, you don't allow any type of Object.

kevmoo:
> tests pass for me locally....

Grrr. I'll rebase once https://github.com/dart-lang/source_gen/pull/342 lands and see what happens

source_gen/lib/src/output_helpers.dart
jakemac53: kk

source_gen/lib/src/output_helpers.dart
jakemac53: I mean, I get where you are going with that but it still makes the code harder to read imo. I don't care _that_ much though.

If only we had:

```
if (value is! Stream) {
  ...
} else {
  value is now a stream here...
} 
``

Comment (kevmoo) [reply source_gen/lib/src/output_helpers.dart]: agreed!

File: source_gen/CHANGELOG.md (hunk 3 above)
Comment (jakemac53) [reply source_gen/CHANGELOG.md]: sure I don't think that really promotes good patterns, but :man_shrugging:

File: source_gen/lib/src/generator_for_annotation.dart [hunk 8]
```diff
@@ -34,18 +35,37 @@ abstract class GeneratorForAnnotation<T> extends Generator {
...
+    }).fold(new Set<String>(), (Set<String> uniqueValues, value) {
+      value = value?.trim();
+
+      if (value != null && value.isNotEmpty) {
+        uniqueValues.add(value);
+      }
+      return uniqueValues;
+    });
+
+    return values.join('\n\n');
```
natebosch: why double newline?

source_gen/lib/src/output_helpers.dart
```diff
@@ -0,0 +1,64 @@
...
+
+      throw _argError(e);
+    });
+  }
+  throw _argError(value);
+}
+
+/// `Stream.fromIterable` has some weird behavior with errors. See
+/// See https://github.com/dart-lang/sdk/issues/33431
+Stream safeIteratorToStream(Iterable source) {
```
natebosch: [nit] This is `safeIterableToStream` not `safeIteratorToStream`. `Iterator` and `Iterable` have different meanings. We use an `Iterator` internal to the method but that's not what we take in.

source_gen/lib/src/output_helpers.dart
```diff
@@ -0,0 +1,64 @@
...
+  var controller = new StreamController();
```
natebosch: do we know here we're operating on `Iterable<String>` and thus should construct a `StreamController<String>`? Or should the method signature be `Stream<T> safeIterableToSream<T>(Iterable<T> source)`?

source_gen/lib/src/output_helpers.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  new Future(() {
```
natebosch: why is this in a `Future`?

source_gen/lib/src/generator_for_annotation.dart
```diff
@@ -34,18 +35,37 @@ abstract class GeneratorForAnnotation<T> extends Generator {
...
   }
 
   /// Override to return source code to generate for [element].
   ///
   /// This method is invoked based on finding elements annotated with an
   /// instance of [T]. The [annotation] is provided as a [ConstantReader].
-  FutureOr<String> generateForAnnotatedElement(
+  ///
+  /// Supported return values include a single [String] or multiple [String]
```
natebosch: we need this comment to explain the deduping.

source_gen/lib/src/generator_for_annotation.dart
kevmoo: So there is whitespace between items. Otherwise back-to-back fields will be glued together, for example.

If that's what a generator wants, it can emit multiple fields together in one chunk.

source_gen/lib/src/output_helpers.dart
kevmoo: I'm nixing all of this, since I don't care about streaming the output all the way to _Builder...stay tuned

source_gen/lib/src/output_helpers.dart
kevmoo: See above – nixing...

Turns out it doesn't matter. Could/should be <Object> if anything

source_gen/lib/src/output_helpers.dart
kevmoo: I was trying to be clever. I didn't want to delay the return of the stream for all of the results to be iterated.

Turns out, they are all iterated and buffered even WITH this. PR pending to remove this.

I've learned a lot 😄

source_gen/lib/src/generator_for_annotation.dart
kevmoo: ack PR coming

PR #342, opened 2018-06-13
CombiningBuilder

grouma:
Towards https://github.com/dart-lang/source_gen/issues/319


source_gen/lib/builders.dart
```diff
@@ -0,0 +1,32 @@
...
+import 'package:glob/glob.dart';
+import 'src/builder.dart';
+
+const _outputExtensions = '.g.dart';
+const _partFiles = '.g.part';
+
+/// A [Builder] which combines part files generated from [SharedPartBuilder].
+class CombiningBuilder extends Builder {
+  @override
+  final Map<String, List<String>> buildExtensions = {
```
natebosch: need a `const` here too if we want it to be fully const

source_gen/lib/builders.dart
```diff
@@ -0,0 +1,32 @@
...
+import 'dart:async';
+
+import 'package:build/build.dart';
+import 'package:glob/glob.dart';
+import 'src/builder.dart';
+
+const _outputExtensions = '.g.dart';
+const _partFiles = '.g.part';
+
+/// A [Builder] which combines part files generated from [SharedPartBuilder].
```
natebosch: worth calling out the pattern it will glob in the doc comment.

source_gen/lib/src/builder.dart
```diff
@@ -144,6 +144,34 @@ class _Builder extends Builder {
   }
 }
 
+/// A [Builder] which generates `part of` files.
```
natebosch: expand this

source_gen/lib/src/builder.dart
```diff
@@ -144,6 +144,34 @@ class _Builder extends Builder {
...
+class SharedPartBuilder extends _Builder {
+  /// Wrap [generators] as a [Builder] that generates `part of` files.
+  ///
+  /// [generatedExtension] indicates what files will be created for each `.dart`
```
natebosch: let's pick a different name for this argument. Maybe `partId` or similar.

source_gen/lib/builders.dart
```diff
@@ -0,0 +1,34 @@
...
+  @override
+  final Map<String, List<String>> buildExtensions = const {
+    '.dart': const [_outputExtensions]
+  };
+
+  @override
+  Future build(BuildStep buildStep) async {
+    var assets = await Future.wait(await buildStep
+        .findAssets(
+            new Glob(buildStep.inputId.changeExtension('.*$_partFiles').path))
```
natebosch: consider pulling out a `pattern` variable so this formats nicer.

If you want to add a dependency on `stream_transform` this chunk can also be rewritten to avoid the nested `await` calls.

```dart
    var pattern = buildStep.inputId.changeExtension('.*$_partFiles').path;
    var combined = await buildStep
        .findAssets(new Glob(pattern))
        .transform(concurrentAsyncMap(buildStep.readAsString))
        .join('\n');
```

Then `buildStep.writeAsString(outputId, combined);`.

source_gen/lib/builders.dart
```diff
@@ -0,0 +1,34 @@
...
+
+import 'dart:async';
+
+import 'package:build/build.dart';
+import 'package:glob/glob.dart';
+import 'src/builder.dart';
+
+const _outputExtensions = '.g.dart';
+const _partFiles = '.g.part';
+
```
natebosch: we'll eventually need a `BuilderFactory` here too. Depending on how you want to structure the tests you could make the class private or skip the `export` for it since all that's needed to be public is the factory.

source_gen/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
   * These changes will likely have no effect on output, unless you customize
     the code formatter.
 
+* Introduce `SharedPartBuilder` for creating part files that can be merged
```
natebosch: I think probably we'll want to make a breaking change in the existing `PartBuilder` to make the extension argument required. I think we probably want everything to go out together. Should we drop a release of the current stuff before adding this?

source_gen/lib/builders.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:build/build.dart';
+import 'package:glob/glob.dart';
+import 'src/builder.dart';
```
natebosch: blank line above this.

source_gen/test/builder_test.dart
```diff
@@ -198,6 +199,80 @@ void main() {
...
+  });
+
+  test('SharedPartBuilder outputs `part of` files', () async {
+    await testBuilder(
+        new SharedPartBuilder([const UnformattedCodeGenerator()], '.foo',
+            header: ''),
+        {'$_pkgName|lib/a.dart': 'library a; part "a.g.dart";'},
+        generateFor: new Set.from(['$_pkgName|lib/a.dart']),
+        outputs: {
+          '$_pkgName|lib/a.foo.g.part': decodedMatches(contains('part of')),
```
natebosch: hmm, we have more to do here. we definitely _don't_ want the shared part builder outputting `part of` statements, and I guess we'll need to add one to the `CombiningBuilder` output.

grouma:
Should be good to go now. PTAL.

source_gen/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.9.0
+
+* `PartBuilder` no longer takes configurable generated outputs. It will only
```
kevmoo: This does not match the chat we just had.

source_gen/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+  output to `.g.dart`.
+* Introduce `SharedPartBuilder` for creating part files that can be merged
```
kevmoo: CombiningBuilder *always* does `.g.dart`

source_gen/lib/builders.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  @override
+  Future build(BuildStep buildStep) async {
+    var pattern = buildStep.inputId.changeExtension('.*$_partFiles').path;
+    var assets = await buildStep
+        .findAssets(new Glob(pattern))
+        .transform(concurrentAsyncMap(buildStep.readAsString))
+        .join('\n');
+    if (assets.isEmpty) return;
+    var output = 'part of\n$assets';
```
natebosch: needs to be (I think) `part of '${buildStep.inputId.uri}';\n$assets`

source_gen/lib/src/builder.dart
```diff
@@ -30,6 +30,9 @@ class _Builder extends Builder {
 
   final String _header;
 
+  /// Whether to include `part of` in the output.
+  final bool _outputPartOf;
```
natebosch: move this next to `_isStandalone` and document explicitly how they differ

kevmoo:
@grouma – I'm digging in on the travis failure.

source_gen/lib/src/builder.dart
```diff
@@ -144,24 +152,57 @@ class _Builder extends Builder {
...
+  /// [formatOutput] is called to format the generated code. Defaults to
+  /// [DartFormatter.format].
+  ///
+  /// [header] is used to specify the content at the top of each generated file.
+  /// If `null`, the content of [defaultFileHeader] is used.
+  /// If [header] is an empty `String` no header is added.
+  SharedPartBuilder(List<Generator> generators, String partId,
+      {String formatOutput(String code),
+      List<String> additionalOutputExtensions = const [],
+      String header})
```
natebosch: I don't think header is something we can/want to support with this

@kevmoo - are you OK if we drop this?

kevmoo: We need to keep it – it allows us to support copyright headers, `// ignore_for_file`, etc...

jakemac53: Move it to the combining builder?

jakemac53: Do we have any use cases where a builder itself is configuring this? I thought this was always configured by the user via build.yaml.

grouma:
The `CombiningBuilder` now takes an optional `header` argument. I believe this should satisfy @kevmoo needs. PTAL.

source_gen/lib/src/builder.dart
```diff
@@ -144,24 +152,51 @@ class _Builder extends Builder {
...
+  /// in [additionalOutputExtensions].
+  ///
+  /// [formatOutput] is called to format the generated code. Defaults to
+  /// [DartFormatter.format].
+  SharedPartBuilder(List<Generator> generators, String partId,
+      {String formatOutput(String code),
+      List<String> additionalOutputExtensions = const []})
+      : super(generators,
+            formatOutput: formatOutput,
+            generatedExtension: '$partId.g.part',
```
kevmoo: should have validation here for `partId` -> `generatedExtension`.

User is required to provide a `.` before – that's broken. I'd argue that it should be `A-Za-z0-9_\.`with no before/after `.`

source_gen/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## 0.9.0
```
kevmoo: Please add docs to the readme about how to configure this!

grouma:
@kevmoo with your changes this should be good to go. PTAL.

PR #343, opened 2018-06-14
prep for release

grouma:


grouma:
Yes. This will be without those features.

PR #348, opened 2018-06-18
Fixes for CombiningBuilder

kevmoo:
Will land seperately

### DIFF
--- a/source_gen/lib/src/builder.dart
+++ b/source_gen/lib/src/builder.dart
@@ -175,6 +175,7 @@ class SharedPartBuilder extends _Builder {
             formatOutput: formatOutput,
             generatedExtension: '.$partId.g.part',
             additionalOutputExtensions: additionalOutputExtensions,
+            header: '',
             outputPartOf: false) {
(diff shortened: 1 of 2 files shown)


source_gen/lib/src/builder.dart
```diff
@@ -106,13 +106,22 @@ class _Builder extends Builder {
                 'Consider adding the following to your source file:\n\n'
                 'library $suggest;');
       }
-      final part = computePartUrl(buildStep.inputId, outputId);
+      contentBuffer.writeln();
+
+      String part;
+      if (_outputPartOf) {
```
grouma: Discussed offline. These variable names are confusing and not indicative in what we are using them for.

kevmoo: right...but is this PR okay otherwise?

grouma: I was hoping you'd fix it in this one.

kevmoo:
> I was hoping you'd fix it in this one.

I'll do a follow-up.

PR #349, opened 2018-06-18
cleanup: simplify _Builder logic around its subtypes

kevmoo:


source_gen/lib/src/builder.dart
```diff
@@ -26,13 +26,7 @@ class _Builder extends Builder {
...
 
   /// Whether to emit a standalone (non-`part`) file in this builder.
-  final bool _isStandalone;
-
-  /// Whether to include `part of` in the output.
-  ///
-  /// This allows the easy merging of outputs without multiple `part of`
-  /// statements.
-  final bool _outputPartOf;
+  bool get _isLibraryBuilder => this is LibraryBuilder;
```
natebosch: 😢

kevmoo:
We were right!

PR #354, opened 2018-06-19
feature(CombiningBuilder): add optional `includePartName` param

kevmoo:
Includes the name of the source parts in a comment before its content
Nice when debugging

### DIFF
--- a/source_gen/lib/builder.dart
+++ b/source_gen/lib/builder.dart
@@ -24,14 +24,19 @@ const _outputExtensions = '.g.dart';
 const _partFiles = '.g.part';
 
 Builder combiningBuilder([BuilderOptions options]) {
-  if (options != null && options.config.isNotEmpty) {
+  var optionsMap = new Map<String, dynamic>.from(options?.config ?? {});
+
+  var builder = new CombiningBuilder(
+      includePartName: optionsMap.remove('include_part_name') as bool);
+
+  if (optionsMap.isNotEmpty) {
     if (log == null) {
       throw new StateError('Upgrade `build_runner` to at least 0.8.2.');
     } else {
-      log.warning('These options were ignored: `${options.config}`.');
+      log.warning('These options were ignored: `$optionsMap`.');
     }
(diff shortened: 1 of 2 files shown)


natebosch:
When do you think we'd want to avoid this? Should we just always have it turned on?

kevmoo:
I think it'd just add noise in the default case – duplicate content from the child builder.

source_gen/lib/builder.dart
```diff
@@ -38,12 +38,20 @@ Builder combiningBuilder([BuilderOptions options]) {
...
         '.dart': const [_outputExtensions]
       };
 
-  const CombiningBuilder();
+  /// Returns a new [CombiningBuilder].
+  ///
+  /// If [includePartName] is `true`, the name of each source part file
+  /// is output as a comment before its content. This can be useful when
+  /// debugging build issues.
+  const CombiningBuilder({bool includePartName = false})
```
natebosch: how would we use this? It's not plumbed through the options

source_gen/lib/builder.dart
```diff
@@ -78,8 +86,13 @@ class CombiningBuilder implements Builder {
     });
 
     var assets = await new Stream.fromIterable(assetIds)
-        .asyncMap(buildStep.readAsString)
-        .map((s) => s.trim())
+        .asyncMap((id) async {
+          var content = (await buildStep.readAsString(id)).trim();
+          if (_includePartName) {
+            content = '// Part: ${id.pathSegments.last}\n$content';
```
natebosch: do we want to omit empty parts?

kevmoo: Nope. This is intentional. Helps folks see that there was a part – just the content was empty.

source_gen/lib/builder.dart
kevmoo: Oops...ack...

kevmoo: PTAL

PR #355, opened 2018-06-20
Drop custom comparison function

natebosch:
`AssetId` implements `Complarable<AssetId>`


jakemac53:
We need to bump the minimum constraint to get this right @natebosch ?

natebosch:
> We need to bump the minimum constraint to get this 

It's been comparable since the beginning, we just didn't bother to check whether it was when writing this 😝

PR #378, opened 2018-09-11
Stop using mapMap from package:collection

natebosch:
The new method on Map directly is nicer.

Also clean up trailing whitespace.

### DIFF
--- a/source_gen/test/constants_test.dart
+++ b/source_gen/test/constants_test.dart
@@ -2,9 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'package:analyzer/dart/constant/value.dart';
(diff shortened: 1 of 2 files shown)


source_gen/test/constants_test.dart
```diff
@@ -2,9 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'package:analyzer/dart/constant/value.dart';
```
kevmoo: any clue what this was used for?

natebosch: There was an explicit generic type parameter on line 127 that is no longer necessary because of better type inference

kevmoo: sweet!

kevmoo:
Land it!

Issue #379, opened 2018-09-18
InvalidGenerationSourceError.toString() method may throw


samogot:
`InvalidGenerationSourceError.toString()` method may throw, and that ruin whole reason of this error.

See https://github.com/dart-lang/json_serializable/issues/315 for details

Although root problem in this case is probably in `analyzer` package or whatever else initializes  `nameOffset` as `-1`, it might make sense to catch possible errors around `spanForElement` call and output only file name if exact span is unavailable for whatever reason.

samogot:
CC @jakemac53

PR #382, opened 2018-10-03
Correctly handle properties in spanForElement, prepare for release

kevmoo:
Fixes https://github.com/dart-lang/source_gen/issues/379

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.9.1+2
+
+* Correctly handle properties in `spanForElement`.
+
 ## 0.9.1+1
 
 * Allow `package:build` version 1.0.0.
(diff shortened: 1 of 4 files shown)


kevmoo:
CC @samogot

samogot:
Do I understand it correctly, that the problem was not in overriding, but it using getter instead of field?

kevmoo:
> Do I understand it correctly, that the problem was not in overriding, but it using getter instead of field?

Yup. Properties are synthetic – they don't have a real location – you have to dig into either the getter or teh setter.

samogot:
Will this code work properly if we have both getter and setter with the same name? Or it will always highlight getter even if the error is related to setter?

kevmoo:
If the error is specific to the setter in a getter/setter pair, the calling code will need to be sure to pass in the setter specifically.

If it's vague, you're right, the getter is picked.

PR #384, opened 2018-10-20
Remove some analyzer ast and src imports

natebosch:
- Use `LibraryElement.exportNamespace` to find types across a libraries
entire visible surface area rather than referencing a
`NamespaceBuilder`.
- Use the different getters for element typs from
`CompilationUnitElement` to get all members of a library rather than
getting the declarations from a `CompilationUnit`.
- Increase minimum anlayzer version to ensure that the `mixins` getter
is available.

### DIFF
--- a/source_gen/lib/src/library.dart
+++ b/source_gen/lib/src/library.dart
@@ -2,11 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'package:analyzer/dart/ast/ast.dart';
-import 'package:analyzer/dart/ast/standard_resolution_map.dart';
 import 'package:analyzer/dart/element/element.dart';
-// ignore: implementation_imports
-import 'package:analyzer/src/dart/resolver/scope.dart';
 import 'package:build/build.dart';
 import 'package:path/path.dart' as p;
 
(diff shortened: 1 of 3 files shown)


source_gen/lib/src/library.dart
```diff
@@ -26,26 +22,27 @@ class AnnotatedElement {
...
+    return type is ClassElement ? type : null;
+  }
 
   /// All of the declarations in this library.
   Iterable<Element> get allElements sync* {
     for (var cu in element.units) {
-      for (var compUnitMember in cu.unit.declarations) {
-        yield* _getElements(compUnitMember);
-      }
+      yield* cu.accessors;
```
kevmoo: It'd be GREAT if analyzer had an API for this ( CC @scheglov @bwilkerson ) – could we get a TODO in here w/ a feature request?

These are places where things fall apart for weird reasons if things are reclassified, etc...

source_gen/pubspec.yaml
```diff
@@ -1,12 +1,12 @@
 name: source_gen
-version: 0.9.1+3
+version: 0.9.2-dev
 author: Dart Team <misc@dartlang.org>
 description: Automated source code generation for Dart.
 homepage: https://github.com/dart-lang/source_gen
 environment:
   sdk: '>=2.0.0-dev.64.0 <3.0.0'
```
kevmoo: Might as well drop 2.0.0-dev version support while you're at it...

source_gen/lib/src/library.dart
```diff
@@ -2,11 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'package:analyzer/dart/ast/ast.dart';
-import 'package:analyzer/dart/ast/standard_resolution_map.dart';
 import 'package:analyzer/dart/element/element.dart';
-// ignore: implementation_imports
-import 'package:analyzer/src/dart/resolver/scope.dart';
```
kevmoo: Nice to get rid of `src` imports!

source_gen/lib/src/library.dart
```diff
@@ -26,26 +22,27 @@ class AnnotatedElement {
...
-  ClassElement findType(String name) =>
-      element.getType(name) ?? _namespace.get(name) as ClassElement;
+  ClassElement findType(String name) {
+    var type = element.exportNamespace.get(name);
+    return type is ClassElement ? type : null;
+  }
 
   /// All of the declarations in this library.
   Iterable<Element> get allElements sync* {
     for (var cu in element.units) {
```
jakemac53: So `units` is fine but `unit` isn't?

bwilkerson:
Yes. Members that access elements from the element model are safe; members
that access the AST from the element model are not.

PR #386, opened 2018-12-05
dont create invalid parts, and allow analyzer 0.34.x

jakemac53:
also prepare to release 0.9.2

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 0.9.2
 
 * Avoid using the AST analyzer model from `LibraryReader`.
+* Don't create part files with no corresponding library.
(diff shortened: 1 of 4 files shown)


source_gen/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## 0.9.2
 
 * Avoid using the AST analyzer model from `LibraryReader`.
+* Don't create part files with no corresponding library.
```
natebosch: are we worried about the potential bad interaction with already published versions of `build_resolvers` where the part directives will appear to be missing?

jakemac53: No - the build_resolvers package was already doing the right thing actually (returning a non-existing  source instead of null)

PR #389, opened 2018-12-12
Add LibraryReader.enums

### LINKED ISSUE #388: LibraryReader.classElements should include enums
The iterable obtained from `LibraryReader.classElements` does not contain enum classes defined in the respective compilation unit. Given that enums are also represented with a `ClassElement` from the analyzer, I would expect them to be included.
I see that the documentation specifically points out that `class` elements defined in the library will be returned, but I think that this makes the name `classElements` a bit confusing. Perhaps there should be a getter `classes`, one for `enums` and one for both?

natebosch:
Fixes #388

Deprecate the `classElements` getter since it is ambiguous, both classes
and enums are represented by `ClassElement` in analyzer, but we don't
want to exclusively always mix them in this interface. Add the name
`classes` to be more specific. Add an implementation for `enums.`

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.9.3
 
-* Rename `LibraryReader.classElemetns` to `LibraryReader.classes` to better
+* Rename `LibraryReader.classElements` to `LibraryReader.classes` to better
   reflect the behavior. Deprecate the old name.
 * Add `LibraryReader.enums`.
 
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.9.3
+
+* Rename `LibraryReader.classElemetns` to `LibraryReader.classes` to better
+  reflect the behavior. Deprecate the old name.
+* Add `LibraryReader.enums`.
+
 ## 0.9.2
 
 * Avoid using the AST analyzer model from `LibraryReader`.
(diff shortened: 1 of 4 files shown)


source_gen/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.9.3
+
+* Rename `LibraryReader.classElemetns` to `LibraryReader.classes` to better
```
jakemac53: Elements :)

source_gen/lib/src/library.dart
```diff
@@ -161,7 +161,12 @@ class LibraryReader {
...
-  /// All of the `class` elements in this library.
-  Iterable<ClassElement> get classElements =>
-      element.units.expand((cu) => cu.types);
+  /// All of the elements reperesenting classes in this library.
+  Iterable<ClassElement> get classes => element.units.expand((cu) => cu.types);
+
+  @Deprecated('Use classes instead')
+  Iterable<ClassElement> get classElements => classes;
+
+  /// All of the elemetns representing enums in this library.
```
jakemac53: elements :)

source_gen/test/library/find_type_test.dart
```diff
@@ -36,9 +39,15 @@ void main() {
   });
 
   test('class count', () {
+    expect(library.classes.map((c) => c.name), ['Example', 'PartClass']);
+    // ignore: deprecated_member_use
     expect(library.classElements, hasLength(2));
```
jakemac53: nit: check its equal to `library.classes`?

source_gen/CHANGELOG.md
natebosch: Done

Issue #322, opened 2018-04-05
Explain how to use this package


Rodsevich:
I've been trying hard enough to figure out how to use this package, reading all the extense build package documentation and that... and I'm still lost.
Can someone please put some trivial e2e example in the wiki, with both the code and the commands needed to generate code and execute it in your project?
Thanks in advance

kevmoo:
Here are two examples:

https://github.com/dart-lang/json_serializable/tree/master/json_serializable
https://github.com/kevmoo/build_cli/tree/master/build_cli

`build_cli` is more simple, so it might be a good spot to start.

Agreed, we need better docs here, but this may get you unblocked...

matanlurey:
I think @kevmoo's comments are likely all we'll have at this point.

Pull requests welcome!

kevmoo:
I'm going to reopen and keep this assigned to me. Once we publish the bits on master, I want to come back here and add an example.

kevmoo:
I have a PR out w/ some initial examples. https://github.com/dart-lang/source_gen/pull/390

graphicbeacon:
Hi @kevmoo, thanks for putting up the examples. After spending hours looking over the code, its beginning to click.

However I've seen in `example_usage/build.yaml` this build configuration for **SharedPartBuilders**:

```yaml
source_gen_example|property_sum:
  generate_for:
    - lib/*.dart
source_gen_example|property_product:
  generate_for:
    - lib/*.dart
```
And I would like to know: 

**Why upon removing this configuration, the contents of the build output file `example_usage/lib/library_source.g.dart` remain the same?**

So far I've seen that it relates to this line in `example_usage/lib/library_source.dart`:

```dart
part 'library_source.g.dart';
```

Removing that line understandably does not create `library_source.g.dart`, **so what then is the purpose of the snippet in `example_usage/build.yaml`**?

natebosch:
@graphicbeacon - the purpose of the `generate_for` argument in the example is to _limit_ the files that are generated. By default a Builder can apply to all files in a target, and by default the target includes everything in `lib/`, `web/`, `bin/`, and `example/` among others:

https://github.com/dart-lang/build/blob/92a4482a8407009d36f5c94601c4acdd58b64569/build_runner_core/lib/src/generate/options.dart#L21-L31

[...]

PR #390, opened 2018-12-21
Examples: add trivial Library and SharedPart builder examples

kevmoo:
Progress towards #322


example/build.yaml
```diff
@@ -0,0 +1,30 @@
...
+builders:
+  # name of the builder
+  member_count:
+    # library URI containing the builder - maps to `lib/member_count_library_generator.dart`
+    import: "package:source_gen_example/builder.dart"
+    # Name of the function in the above library to call.
+    builder_factories: ["metadataLibraryBuilder"]
+    # The mapping from the source extension to the generated file extension
+    build_extensions: {".dart": [".info.dart"]}
+    # Will automatically run on any package that depends on it
```
natebosch: [nit] Just as a note, this will only run on the _root_ package if it depends on it because of the `build_to: source`. I'm not sure if that's worth calling out here or not.

kevmoo: I thought that was implied. No?

natebosch: 🤷‍♂️ 

I guess it's implied, but as I read this I was thrown off by the "any package" here.

PR #396, opened 2019-01-14
Use ParsedLibraryImpl to find source span

natebosch:
The old code path does not work when the ElementAnnotation was returned
by an `AnalysisDriver` instead `AnalysisContext`. Prepare for
`build_resolvers` to switch to `AnalysisDriver` by migrating to a
forwards compatible approach.


natebosch:
Thanks to @scheglov for the original patch

PR #404, opened 2019-02-28
Add source_gen_test package

### LINKED ISSUE #403: Extract json_serializable test utils into an independant package
An unusual request for once:

I've been exploring a bit into the test folder of json_serializable and I got quite surprised to find a small testing framework included.

I think it would be interesting to extract everything related to `ShouldThrow`/`ShouldGenerate` into a reusable package.

kevmoo:
Fixes #403

### DIFF
--- a/source_gen_test/lib/src/annotations.dart
+++ b/source_gen_test/lib/src/annotations.dart
@@ -54,12 +54,17 @@ class ShouldThrow extends TestExpectation {
   final String errorMessage;
   final String todo;
 
+  /// Defaults to `true`.
+  final bool elementShouldMatchAnnotated;
+
   const ShouldThrow(
     this.errorMessage, {
     this.todo,
+    bool elementShouldMatchAnnotated = true,
     Iterable<String> configurations,
     List<String> expectedLogItems,
-  }) : super._(configurations, expectedLogItems);
+  })  : elementShouldMatchAnnotated = elementShouldMatchAnnotated ?? true,
+        super._(configurations, expectedLogItems);
 
   @override
   TestExpectation replaceConfiguration(Iterable<String> newConfiguration) {
--- a/source_gen_test/lib/src/annotations.dart
+++ b/source_gen_test/lib/src/annotations.dart
@@ -78,9 +78,10 @@ class ShouldThrow extends TestExpectation {
     assert(newConfiguration != null);
     return ShouldThrow(
       errorMessage,
-      todo: todo,
       configurations: newConfiguration,
+      element: element,
       expectedLogItems: expectedLogItems,
+      todo: todo,
     );
   }
 }
--- a/source_gen_test/lib/src/annotations.dart
+++ b/source_gen_test/lib/src/annotations.dart
@@ -1,0 +1,63 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Non-public, implementation base class of  [ShouldGenerate] and
+/// [ShouldThrow].
+abstract class TestExpectation {
+  final Iterable<String> configurations;
+  final List<String> expectedLogItems;
+
+  const TestExpectation._(this.configurations, List<String> expectedLogItems)
+      : expectedLogItems = expectedLogItems ?? const [];
+
+  TestExpectation replaceConfiguration(Iterable<String> newConfiguration);
+}
+
+const defaultConfigurationName = 'default';
+
+class ShouldGenerate extends TestExpectation {
+  final String expectedOutput;
+  final bool contains;
+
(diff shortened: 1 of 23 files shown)


rrousselGit:
Looks great!

I'm not sure that using annotations is the best approach though. I can see a few drawbacks:

- It modifies the code we want to test which may be undesired. For example my `functional_widget` transpose annotations from the decorated element to the generated one.

- It ties the source to only one output. Based on configurations, it is definitely possible for one source to have different outputs. `functional_widget` is again an example here, where based on the `build.yaml` it may generate more stuff.

- We're loosing on readability a bit. It's quite unusual and make it a bit harder for a non-initiated to find all the related pieces.

____

What I think makes more sense is to have something similar to Flutter's `tester.pumpWidget`. 

When I ported your work to `functional_widget`, I tried to make one of these and I think it makes things clearer/more flexible (Although I'm not a fan of my API either)

A random though: Would a custom `Matcher` make sense?

```dart
await expectLater(
  SomeGenerator(),
  generatesFor(
    'path.dart': 
    '''output''',
    finder.byName('name'),
  ),
);
```

Where the 3rd argument is optional and allows selecting 1+ elements to run the generator on (but only if the generator is a `GeneratorForAnnotation`).

kevmoo:
> Looks great!
> 
> I'm not sure that using annotations is the best approach though. I can see a few drawbacks:
> 
> * It modifies the code we want to test which may be undesired. For example my `functional_widget` transpose annotations from the decorated element to the generated one.

How? I'm not sure I follow.

> * It ties the source to only one output. Based on configurations, it is definitely possible for one source to have different outputs. `functional_widget` is again an example here, where based on the `build.yaml` it may generate more stuff.

You can provide multiple configurations and test them.

> * We're loosing on readability a bit. It's quite unusual and make it a bit harder for a non-initiated to find all the related pieces.

I think keeping the source and expected output together eases maintenance. You don't have to use the annotation syntax, though. You can just use `generateForElement` by itself.

> What I think makes more sense is to have something similar to Flutter's `tester.pumpWidget`.

I'm not familiar with that...

> A random though: Would a custom `Matcher` make sense?

Not a bad idea! Want to try and build that after I land this?

rrousselGit:
> You can provide multiple configurations and test them.

Do you have an example? The other two points I made are not that important, but I think this one is.

How would `@ShouldGenerate` works if the output is based on the generator configurations?

> Not a bad idea! Want to try and build that after I land this?

If I find the time to, then sure! It's not my priority though, as I have a workaround.

kevmoo:
> > You can provide multiple configurations and test them.
> 
> Do you have an example? The other two points I made are not that important, but I think this one is.

```dart
@ShouldThrow(
  'All classes must start with `TestClass`.',
  todo: 'Rename the type or remove the `TestAnnotation` from class.',
  configurations: ['default'],
  expectedLogItems: ['This member might be not good.'],
)
@ShouldGenerate(
  r'''
const BadTestClassNameLength = 12;

const BadTestClassNameLowerCase = badtestclass;
''',
  configurations: ['no-prefix-required'],
  expectedLogItems: ['This member might be not good.'],
)
@ShouldThrow(
  'Uh...',
  configurations: ['vague'],
  elementShouldMatchAnnotated: false,
)
@TestAnnotation()
class BadTestClass {}
```

> How would `@ShouldGenerate` works if the output is based on the generator configurations?

You provide `additionalGenerators` and specify alternate `ShouldGenerate` annotations as appropriate.

> > Not a bad idea! Want to try and build that after I land this?
> 
> If I find the time to, then sure! It's not my priority though, as I have a workaround.

Cool.

rrousselGit:
Oh I see! I missed that part, my bad. It's fine then. 😄 

It would be cool to have some examples somewhere in the repo (in the `example` folder?)

kevmoo:
> Oh I see! I missed that part, my bad. It's fine then. 😄

Glad you approve!

> It would be cool to have some examples somewhere in the repo (in the `example` folder?)

I'll get there. 😉

source_gen_test/lib/src/annotations.dart
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore_for_file: comment_references
```
natebosch: get rid of the references. They aren't worth putting in if they don't work.

kevmoo: with special hacks in `dartdoc` they often DO work.

kevmoo:
@natebosch – I'll put this in my own repo for now and we can talk about migrating it.

@rrousselGit – don't worry. I'll CC you on my new repo.

kevmoo:
@rrousselGit see here - https://github.com/kevmoo/source_gen_test

PR #405, opened 2019-02-28
2 commits: lints and test on Dart 2.0.0

kevmoo:


kevmoo:
@jakemac53 – mind taking a look since the other guys are OOO?

example/mono_pkg.yaml
```diff
@@ -2,6 +2,7 @@
 dart:
   - dev
   - stable
```
jakemac53: Consider removing stable?

source_gen/lib/src/generator_for_annotation.dart
```diff
@@ -66,6 +66,6 @@ abstract class GeneratorForAnnotation<T> extends Generator {
   ///
   /// Implementations should return `null` when no content is generated. Empty
   /// or whitespace-only [String] instances are also ignored.
-  generateForAnnotatedElement(
+  dynamic generateForAnnotatedElement(
```
jakemac53: Shouldn't this really be `FutureOr<String>` also?

example/mono_pkg.yaml
kevmoo: soon enough we'll have 2.3.0-dev, and we'll want this back

source_gen/lib/src/generator_for_annotation.dart
kevmoo: See the doc comment. This can return an iterable or Stream, too...

example/mono_pkg.yaml
jakemac53: why? Testing the minimum (2.0.0) and maximum (dev) should be sufficient.

Maybe current stable on the cron job or something, but not every PR imo.

kevmoo: Ah...decent point.

PR #407, opened 2019-03-01
Add GeneratorForAnnotation.generateForAnnotatedElementStream

kevmoo:


kevmoo:
towards https://github.com/dart-lang/source_gen/issues/403

natebosch:
Why is this useful? `generateForAnnotatedElement` can return a `Stream<String>`.

kevmoo:
> Why is this useful? `generateForAnnotatedElement` can return a `Stream<String>`.

This normalizes the return value (which can be many things) to `Stream<String>` – it's mostly for testing generators.

natebosch:
>  it's mostly for testing generators.

I'm not in favor of increasing the surface area and cognitive load for implementors to do that - if we do decide to move forward with a test package we can copy/paste the normalize function into that package.

kevmoo:
> I'm not in favor of increasing the surface area and cognitive load for implementors to do that - if we do decide to move forward with a test package we can copy/paste the normalize function into that package.

ack

PR #413, opened 2019-04-03
Migrate off ParsedLibraryResultImpl

natebosch:
This was a transition API until we could know that `element.session`
would be available. With analyzer above version `0.35.0` we know that
`build_resolvers` will be using a code path that makes `.session`
available.

Also:
- Remove unnecessarily low minimum dependency versions for packages that
  we would not be able to solve with due to the analyzer constraint.
- Replace an `// ignore: unawaited_futures` with `unawaited` call from
  `package:pedantic`.
- Remove comment referencing closed SDK issue. GOing through the
  `AnnotatedNode` is the correct approach for finding source ranges.
- Rename method to avoid a prefix of `get`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,68 +1,68 @@
-# Created with package:mono_repo v1.2.1
+# Created with package:mono_repo v2.0.0
 language: dart
 
 jobs:
   include:
     - stage: analyze_format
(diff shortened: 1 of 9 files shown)


kevmoo:
Need to bump the min SDK for source_gen (and travis)

PR #426, opened 2019-08-19
Ignore deprecated API usage in pkg:analyzer

kevmoo:


example/lib/src/utils.dart
```diff
@@ -8,6 +8,9 @@ import 'package:source_gen/source_gen.dart';
 /// Returns all [TopLevelVariableElement] members in [reader]'s library that
 /// have a type of [num].
 Iterable<TopLevelVariableElement> topLevelNumVariables(LibraryReader reader) =>
-    reader.allElements.whereType<TopLevelVariableElement>().where((element) =>
-        element.type
+    reader.allElements
+        .whereType<TopLevelVariableElement>()
+        .where((element) => element.type
+            // TODO: migrate to supported API when min pkg:analyzer is bumped
```
natebosch: Maybe file an issue and reference it here?

kevmoo: I figured once we want to roll analyzer to a version where this API is gone, we can revisit.

PR #430, opened 2019-09-24
Get element of DartObject using also toFunctionValue().

scheglov:
See https://github.com/dart-lang/sdk/issues/38546#issuecomment-534671075 for the context.

### DIFF
--- a/source_gen/lib/src/constants/revive.dart
+++ b/source_gen/lib/src/constants/revive.dart
@@ -19,8 +19,8 @@ import '../utils.dart';
 /// Dart source code (such as referencing private constructors). It is up to the
 /// build tool(s) using this library to surface error messages to the user.
 Revivable reviveInstance(DartObject object, [LibraryElement origin]) {
-  origin ??= object.type.element.library;
-  final element = object.type.element;
+  final element = object.type.element ?? object.toFunctionValue();
+  origin ??= element.library;
   var url = Uri.parse(urlOfElement(element));
   if (element is FunctionElement) {
     return Revivable._(
--- a/source_gen/pubspec.yaml
+++ b/source_gen/pubspec.yaml
@@ -8,7 +8,7 @@ environment:
   sdk: ">=2.1.0 <3.0.0"
 
 dependencies:
-  analyzer: '>=0.35.0 <0.39.0'
+  analyzer: '>=0.37.1 <0.39.0'
   async: ^2.0.7
   build: ^1.0.0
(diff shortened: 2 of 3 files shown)


source_gen/lib/src/constants/revive.dart
```diff
@@ -19,8 +19,8 @@ import '../utils.dart';
 /// Dart source code (such as referencing private constructors). It is up to the
 /// build tool(s) using this library to surface error messages to the user.
 Revivable reviveInstance(DartObject object, [LibraryElement origin]) {
-  origin ??= object.type.element.library;
-  final element = object.type.element;
+  final element = object.type.element ?? object.toFunctionValue();
```
natebosch: what version of `analyzer` do we need to ensure that `DartObject` has a `toFunctionValue()` method? it looks like it's missing on travis

scheglov: It was added in https://github.com/dart-lang/sdk/commit/4c08c5026d29784099019b68a2f4e6e655d63ab3, so looks like `0.37.1` (which has been published since then).

scheglov: Should I move the lower bound for the `analyzer` dependency version range in the same PR, or do you prefer a separate one?

natebosch: Same PR is fine.

scheglov: Travis is still not happy :-(

srawlins: Looks like https://github.com/dart-lang/source_gen/blob/master/.travis.yml needs to drop support for 2.1.0.

https://github.com/dart-lang/sdk/commit/bb12829571f064174fab0107851a067e0a2b4685#diff-0c4582f82c9ebb7e852f2d4999960c1e

scheglov:
There is a problem with older Dart SDK.
See https://github.com/dart-lang/sdk/issues/38546#issuecomment-534723453

source_gen/pubspec.yaml
```diff
@@ -8,7 +8,7 @@ environment:
   sdk: ">=2.1.0 <3.0.0"
```
natebosch: Do we need to increase the lower bound here?

scheglov: Why?
`package:analyzer` does have a higher lower bound `>=2.2.2 <3.0.0`.
But I don't know what is you motivation to increase.

PR #435, opened 2019-11-06
Fix tests on travis for deprecated API

kevmoo:


kevmoo:
HRM....requires bumping min analyzer to 0.39.0...not sure if we're ready for that...

natebosch:
`build_resolvers` is already at `0.39.0`. I'm fine bumping here as well.

kevmoo:
> `build_resolvers` is already at `0.39.0`. I'm fine bumping here as well.

Wha?

kevmoo:
Otherwise LGTY @natebosch @jakemac53 ?

natebosch:
I don't think it is worth submitting this unless the red travis is blocking something else. We should instead upgrade the version of analyzer.

kevmoo:
> I don't think it is worth submitting this unless the red travis is blocking something else. We should instead upgrade the version of analyzer.

0.40.0 is coming soon. I'd rather get things green (so we catch any other regressions) and wait for that

natebosch:
https://github.com/dart-lang/source_gen/pull/437

PR #436, opened 2019-11-19
fix(typo): util urlOfElement returns valid result

alex-misch:


alex-misch:
> @googlebot I signed it!

PR #437, opened 2019-12-13
Migrate off of deprecated analyzer APIs

natebosch:
`ClassElement.type` is deprecated, use `ClassElement.isntantiate()`
instead.

Bump minimum analyzer to `0.39.0`.


kevmoo:
@natebosch – you'll need to bump the min SDK it seems...

PR #439, opened 2020-01-02
Log more carefully.

davidmorgan:
If analysis is being done using a combination of srcs + summaries, and the element specified is in a summary, then we won't be able to find the src to log it.

Currently this causes an uncaught exception :)

A simple try/catch improves matters.

### DIFF
--- a/source_gen/lib/src/generator.dart
+++ b/source_gen/lib/src/generator.dart
@@ -58,10 +58,9 @@ class InvalidGenerationSourceError extends Error {
           ..writeln(span.start.toolString)
           ..write(span.highlight());
       } catch (_) {
-        buffer
-          ..writeln()
-          ..writeln('While logging this error: '
-              'tried to log source from "$element" but couldn\'t find it.');
+        // Source for `element` wasn't found, it must be in a summary with no
+        // associated source. We can still give the name.
+        buffer..writeln()..writeln('Cause: $element');
       }
     }
 
--- a/source_gen/lib/src/generator.dart
+++ b/source_gen/lib/src/generator.dart
@@ -51,11 +51,18 @@ class InvalidGenerationSourceError extends Error {
     final buffer = StringBuffer(message);
 
     if (element != null) {
-      final span = spanForElement(element);
-      buffer
-        ..writeln()
-        ..writeln(span.start.toolString)
-        ..write(span.highlight());
+      try {
+        final span = spanForElement(element);
+        buffer
+          ..writeln()
+          ..writeln(span.start.toolString)
+          ..write(span.highlight());
+      } catch (_) {
+        buffer
+          ..writeln()
+          ..writeln('While logging this error: '
+              'tried to log source from "$element" but couldn\'t find it.');
+      }
     }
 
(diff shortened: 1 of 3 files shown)


source_gen/lib/src/generator.dart
```diff
@@ -51,11 +51,18 @@ class InvalidGenerationSourceError extends Error {
...
+      try {
+        final span = spanForElement(element);
+        buffer
+          ..writeln()
+          ..writeln(span.start.toolString)
+          ..write(span.highlight());
+      } catch (_) {
+        buffer
+          ..writeln()
+          ..writeln('While logging this error: '
```
jakemac53: The wording here is a bit awkward, maybe `Error finding source span information for "$element".`?

davidmorgan: Yeeeeeah this is a really tough one.

The error message is going to be seen by someone _using_ the generator, who is likely not the generator author. The part of the error message that we already printed contains the main information. Then, the generator author gave us the problem element and we're trying to log that to add more information. I half think it would be better to just log nothing, because the person reading this will likely have no idea what the generator is doing. But I also don't like just completely swallowing the information.

So what I'm aiming for is to give a note that people who don't know the generator will ignore, while still copying+pasting it if they contact the generator author :)

Here's some example output:

```
[SEVERE]: Error running FunkyGenerator
FunkyGenerator failed because you forgot to write "foo bar baz" in your srcs. Sorry. --FunkyGenerator authors.
While logging this error: tried to log source from "class SomeUnavailableDepOfMyClass" but couldn't find it.
```

WDYT?

Thanks :)

jakemac53: Maybe just replace `While logging this error` with `Additional Errors` or something?

davidmorgan: I'm a bit worried they might think the whole problem is caused by SomeUnavailableDepOfMyClass being unavailable, and they'll try to fix the dep on it...when it's actually fine that the src is missing. Maybe less is better:

```
[SEVERE]: Error running FunkyGenerator
FunkyGenerator failed because you forgot to write "foo bar baz" in your srcs. Sorry. --FunkyGenerator authors.
Cause: class SomeUnavailableDepOfMyClass
```

?

jakemac53: That seems fine - its basically just falling back on a less descriptive "source span" at that point.

davidmorgan: SG, done. Thanks!

davidmorgan: (And added a comment).

davidmorgan:
@jakemac53 Thanks! Do you know who usually releases / pulls into google3, please? This fix is a request from an internal user, but not an urgent one.

jakemac53:
One of us would usually do it once we get a nag email about it being out of date :D.

We can do it sooner though if this is meant specifically for internal.

jakemac53:
Can you also update the pubspec/changelog and then we can merge (I would but I can't push to this branch)

davidmorgan:
Done, thanks.

PR #443, opened 2020-01-13
Updates for pkg:analyzer deprecated APIs

kevmoo:


source_gen/lib/src/utils.dart
```diff
@@ -16,15 +16,15 @@ import 'package:path/path.dart' as p;
...
+/// This function will return `'VoidFunc'`, unlike [DartType.getDisplayString].
 String typeNameOf(DartType type) {
   if (type is FunctionType) {
     final element = type.element;
     if (element is GenericFunctionTypeElement) {
       return element.enclosingElement.name;
     }
   }
-  return type.name;
+  return type.getDisplayString();
```
jakemac53: cc @natebosch I thought you did something a bit different here - is there a difference?

natebosch: I did `type.element.name` I believe. I think I did so because it matched the implementation of `type.name`, and because the doc comment on `getDisplayString()` says "Clients should not depend on the content of the returned value as it will be changed if doing so would improve the UX."

kevmoo: PTAL

PR #455, opened 2020-03-11
Streamlined error handling in all builders

kevmoo:
There should be no user-visible behavior change.

### DIFF
--- a/source_gen/lib/src/builder.dart
+++ b/source_gen/lib/src/builder.dart
@@ -126,7 +126,9 @@ class _Builder extends Builder {
         ..writeln()
         ..writeln(_headerLine)
         ..writeAll(
-            LineSplitter.split(item.toString()).map((line) => '// $line\n'))
+          LineSplitter.split(item.generatorDescription)
+              .map((line) => '// $line\n'),
+        )
         ..writeln(_headerLine)
         ..writeln()
         ..writeln(item.output);
--- a/source_gen/lib/src/builder.dart
+++ b/source_gen/lib/src/builder.dart
@@ -289,28 +289,23 @@ Stream<GeneratedOutput> _generate(
   final libraryReader = LibraryReader(library);
   for (var i = 0; i < generators.length; i++) {
     final gen = generators[i];
-    try {
-      var msg = 'Running $gen';
-      if (generators.length > 1) {
-        msg = '$msg - ${i + 1} of ${generators.length}';
-      }
-      log.fine(msg);
-      var createdUnit = await gen.generate(libraryReader, buildStep);
-
-      if (createdUnit == null) {
-        continue;
-      }
-
-      createdUnit = createdUnit.trim();
-      if (createdUnit.isEmpty) {
-        continue;
-      }
-
-      yield GeneratedOutput(gen, createdUnit);
-    } catch (e, stack) {
-      log.severe('Error running $gen', e, stack);
-      yield GeneratedOutput.fromError(gen, e, stack);
+    var msg = 'Running $gen';
+    if (generators.length > 1) {
+      msg = '$msg - ${i + 1} of ${generators.length}';
     }
+    log.fine(msg);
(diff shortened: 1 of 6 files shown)


source_gen/lib/src/generated_output.dart
```diff
@@ -2,29 +2,32 @@
...
 
-import 'dart:convert';
-
 import 'generator.dart';
 
 class GeneratedOutput {
   final String output;
   final Generator generator;
+
+  @Deprecated('Always null. Will be removed in the next release.')
```
natebosch: This class is not in the public API, we don't need to worry about deprecating.

I strongly suspect we can drop this class entirely and yield the `String` directly.

kevmoo: Yup. See follow-up commit

natebosch: Still should remove the class. At this point it is only a wrapper around the `String output`.

kevmoo: Oh...hrm...k

kevmoo: done!

kevmoo:
> FYI you'll have to update some similar tests in google3 to land this. I'll DM you the details.

I don't have a Google3 box at the moment... 🤷‍♂

source_gen/lib/src/builder.dart
```diff
@@ -281,36 +280,31 @@ class LibraryBuilder extends _Builder {
...
-      createdUnit = createdUnit.trim();
-      if (createdUnit.isEmpty) {
-        continue;
-      }
+    if (createdUnit == null) {
+      continue;
+    }
 
-      yield GeneratedOutput(gen, createdUnit);
-    } catch (e, stack) {
```
natebosch: I mentioned this on an internal thread - if we want to keep the behavior of emitting logs for all errors and not just the first one we could keep the try/catch and the `log.severe`.

I don't have strong opinions on whether we'd prefer multiple errors or to fail fast on the first one. @jakemac53 might have an opinion.

For now I'd lean towards this implementation and fail fast.

leonsenft: The issue with logging is that it will potentially swallow build failures intended for the build runner to catch - I'm not sure how we can get both properties.

natebosch: Logging at severe _does_ cause a build failure. The difference are:

 - throw:
   - fail fast
   - error output is based on `toString()` for the exception and not controlled by builder author.
   - May be caught somewhere else in builder author code and ignored (won't be caught and ignored by build system or `source_gen`).
- `log.severe`:
   - fail slow - can emit multiple error messages, but the overall build step will still fail.
   - error output is under your control, not necessary an exception `toString()`.
   - can't be caught and ignored. Every step that logs at severe will be a failure.

kevmoo: catch all is an anti-pattern I'm happy to ditch. Also guessing that the throw case is easier to debug.

kevmoo: I'll just hold off landing this until @jakemac53 can weigh in

leonsenft: Oh right, so whether we log a build failure here or further down in the build runner doesn't actually matter. I was thinking logging it here didn't give the underlying build system a chance to handle it, but as long as that handling takes places on the log record, it's fine.

source_gen/lib/src/generated_output.dart
kevmoo: Hand to revert that last change, @natebosch – we use the generator name in the output.

See my latest commit. I just store the name instead of holding on to the generator.

kevmoo:
@jakemac53 – thoughts?

source_gen/lib/src/generated_output.dart
natebosch: SGTM

jakemac53:
LGTM

PR #469, opened 2020-06-08
Including extensions on allElements

### LINKED ISSUE #468: [LibraryReader] allElements does not include extensions
https://github.com/dart-lang/source_gen/blob/4fb8a656b8108e7ff45b5d1f5048ac6851522854/source_gen/lib/src/library.dart#L37-L59

Since `allElements` does not load extensions if I use `annotatedWith` to get extensions with some annotation it will not return them.

AlbertoMonteiro:
Fixes #468

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Streamlined error handling in all builders. There should be no user-visible
   behavior change.
 * Update minimum Dart SDK to `2.7.0`.
+* Changed the behavior of `LibraryReader.allElements` to include **extensions**
 
 ## 0.9.5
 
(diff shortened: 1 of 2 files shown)


natebosch:
Looks good! Can you add an entry to the CHANGELOG?

kevmoo:
@natebosch – think we need a test here? Also bump the version?

natebosch:
> think we need a test here?

I can take a look at that in a followup.

> Also bump the version?

Version is already bumped appropriately. We are at a `-dev`.

AlbertoMonteiro:
@natebosch I've changed changelog, let me know if need something else 😁

natebosch:
Thanks!

There will be some lag before we can get this published. I've still been fighting some internal tests following #455

AlbertoMonteiro:
Tyvm @natebosch

PR #472, opened 2020-07-02
Include matching Dart language version overrides in generated parts...

kevmoo:
...if the source library has a language version override

Also prepare for v0.9.6


davidmorgan:
LGTM! Tested locally with built_value_generator, it does exactly what is needed: generates nothing if there is no override, matches the override if there is.

Thanks :)

kevmoo:
@natebosch ?

PR #475, opened 2020-07-10
Improve the error message when formatting fails

kevmoo:


source_gen/lib/src/builder.dart
```diff
@@ -139,15 +139,17 @@ class _Builder extends Builder {
...
-      log.severe(
-          'Error formatting generated source code for ${library.identifier} '
-          'which was output to ${outputId.path}.\n'
-          'This may indicate an issue in the generated code or in the '
-          'formatter.\n'
-          'Please check the generated code and file an issue on source_gen if '
-          'appropriate.',
-          e,
-          stack);
+      log.warning(
```
jakemac53, line 142: This is also making it a warning and not failure which seems like the larger change, please indicate that in the changelog and commit message.

kevmoo: Oops! Not intended. I played w/ that idea a bit, but I don't think we want to turn these into silent failures.

Issue #481, opened 2020-09-22
Avoid resolving all inputs when possible


jakemac53:
Version `1.5.0` of the `build` package adds a new api - `compilationUnitFor` - which could be used in order to cheaply skip some libraries that for instance have no annotated elements at all.

The `LibraryReader` would likely need to become lazy - taking only a `BuildStep` + `assetId` and not a `LibraryElement`. In general that could be a good change though.

This has the potential to very dramatically improve the speed of all source_gen builders, which usually today end up fully resolving every single library in any package they are applied to.

natebosch:
How many builders do we think currently don't use the `LibraryReader`?

jakemac53:
This optimization would be pretty specifically targeted at `GeneratorForAnnotation` generators. The goal would be to efficiently skip the vast majority of inputs (at a minimum, those without any top level annotated elements at all which should be most).

I assume once you find a real annotation that you care about most builders would use the `LibraryReader`.

PR #495, opened 2020-12-08
Optimize annotation generators to not resolve libs with no annotations

jakemac53:
Closes https://github.com/dart-lang/source_gen/issues/481 (or as well as I can close that without major breaking changes).

In the case where we have nothing but GeneratorForAnnotation generators, get the compilation unit for the library and if there are zero top level annotations then bail out early.

### DIFF
--- a/source_gen/lib/src/builder.dart
+++ b/source_gen/lib/src/builder.dart
@@ -8,6 +8,7 @@ import 'package:analyzer/dart/element/element.dart';
 import 'package:build/build.dart';
 import 'package:dart_style/dart_style.dart';
 import 'package:pedantic/pedantic.dart';
+import 'package:source_gen/source_gen.dart';
 
 import 'generated_output.dart';
 import 'generator.dart';
@@ -69,7 +70,14 @@ class _Builder extends Builder {
   @override
   Future build(BuildStep buildStep) async {
     final resolver = buildStep.resolver;
+
     if (!await resolver.isLibrary(buildStep.inputId)) return;
+
+    if (_generators.every((g) => g is GeneratorForAnnotation) &&
+        !await _hasAnyTopLevelAnnotations(buildStep.inputId, resolver)) {
+      return;
+    }
+
     final lib = await buildStep.resolver
         .libraryFor(buildStep.inputId, allowSyntaxErrors: allowSyntaxErrors);
     await _generateForLibrary(lib, buildStep);
@@ -332,6 +340,19 @@ Stream<GeneratedOutput> _generate(
 
     yield GeneratedOutput(gen, createdUnit);
   }
+}
+
+Future<bool> _hasAnyTopLevelAnnotations(
+    AssetId input, Resolver resolver) async {
+  final parsed = await resolver.compilationUnitFor(input);
+  final annotatableObjects = [
+    ...parsed.directives,
+    ...parsed.declarations,
+  ];
+  for (var object in annotatableObjects) {
+    if (object.metadata.isNotEmpty) return true;
+  }
+  return false;
 }
 
 final _formatter = DartFormatter();
--- a/source_gen/lib/src/builder.dart
+++ b/source_gen/lib/src/builder.dart
@@ -74,7 +74,7 @@ class _Builder extends Builder {
     if (!await resolver.isLibrary(buildStep.inputId)) return;
 
     if (_generators.every((g) => g is GeneratorForAnnotation) &&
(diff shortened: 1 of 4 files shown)


jakemac53:
cc @lunaticcoding I am curious if this improves your build, and how it compares with the custom build.yaml config (can you try with this version of source_gen but no `build.yaml`?).

You can depend on this like so in your pubspec:

```
dependency_overrides:
  source_gen:
    git:
      url: https://github.com/dart-lang/source_gen.git
      path: source_gen
```

source_gen/lib/src/builder.dart
```diff
@@ -334,6 +342,19 @@ Stream<GeneratedOutput> _generate(
   }
 }
 
+Future<bool> _hasAnyTopLevelAnnotations(
+    AssetId input, Resolver resolver) async {
+  final parsed = await resolver.compilationUnitFor(input);
+  final annotatableObjects = [
```
natebosch, line 348: [nit] I wouldn't collect these to a list before we iterate over them.

Either 2 for loops, or `parsed.directives.followedBy(parsed.declarations)` would avoid an extra list.

source_gen/lib/src/builder.dart
```diff
@@ -69,7 +70,14 @@ class _Builder extends Builder {
   @override
   Future build(BuildStep buildStep) async {
     final resolver = buildStep.resolver;
+
     if (!await resolver.isLibrary(buildStep.inputId)) return;
+
+    if (_generators.every((g) => g is GeneratorForAnnotation) &&
+        !await _hasAnyTopLevelAnnotations(buildStep.inputId, resolver)) {
```
natebosch, line 77: [nit] I find this a bit easier to read if it has parens
```suggestion
        !(await _hasAnyTopLevelAnnotations(buildStep.inputId, resolver))) {
```

source_gen/test/builder_test.dart
```diff
@@ -540,6 +545,17 @@ foo generated content
     expect(builders.toString(),
         'Generating .foo.dart: _LiteralGenerator, _LiteralGenerator');
   });
+
+  test('Does not resolve the library if it doesn\'t need to', () async {
```
natebosch, line 549: [nit] 
```suggestion
  test('Does not resolve the library if there are no top level annotations', () async {
```

natebosch (approved):

source_gen/lib/src/builder.dart
jakemac53: Going with two loops, I can't use `followedBy` without casting the iterable first or creating an empty iterable of the right type followed by the other two both of which were pretty gross.

jakemac53:
Ok, I made an artifical test for this where I applied json_serializable to _all_ packages in the transitive deps of the angular components example package.

That took the build time from 1m4s to 1m46s without this change, and 1m23s with this change. So definitely a large improvement and we should ship this.

I was still surprised that it added so much though, but I think we are still parsing all the files from scratch even if they are already parsed (that was why I wanted to validate this in the first place to ensure we didn't actually make things _worse_). cc @scheglov is that accurate (we use `AnalysisDriver.parseFile` api)?

scheglov:
Yes, when you ask to parse a string, or parse a file, the analyzer will parse it from scratch in both cases.

jakemac53:
> Yes, when you ask to parse a string, or parse a file, the analyzer will parse it from scratch in both cases.

Ok, thanks. Is this only a memory saving optimization, or a deeper limitation? In other words is it feasible to try adding an option to keep these ASTs in memory in the future reasonably?

jakemac53:
@lunaticcoding this is published now so you can just pub upgrade and don't need to add deps, still curious what the results are for you if you have a chance :).

scheglov:
We don't do it because (1) parsing is relatively cheap; (2) clients can cache themselves (the only thing that the client need to track is whether this specific file was changed). I realize that the analyzer should provide some features, and it is not hard to have such cache, but we need to compare adding complexity vs. amount of service.

jakemac53:
I am not sure we can fully cache this ourselves (at least not the AST which was created during resolution)? We could potentially cache the asts that we manually parse, but in this case specifically I want to re-use the AST that was created during the call to `AnalysisDriver#getLibraryByUri`. At the same time I would like calls to `getLibraryByUri` to not re-parse if I have already parsed that file through the `parseFile` api.

This isn't a deal breaker, fwiw, and I don't know how much of the cost I observed is coming from this. Mostly I was wondering if its easy to hack something together to see how much faster it would be if we could do this (and also I could investigate the memory usage of that).

PR #498, opened 2020-12-14
fix bug in early bailout case - check part files for annotations also

jakemac53:
Fixes https://github.com/dart-lang/source_gen/issues/497


kevmoo (approved):
Yay for edge cases!

PR #500, opened 2021-01-21
Allow the library builder to work with non-dart (json, xml, yaml, etc...) based on allow build extension

pratikpparikh:
The current source generator only works with dart code even for LibraryBuilder, which kind not does not make sense. This change will support to generate code for non-dart extension.

### DIFF
--- a/source_gen/lib/src/builder.dart
+++ b/source_gen/lib/src/builder.dart
@@ -73,9 +73,10 @@ class _Builder extends Builder {
   @override
   Future build(BuildStep buildStep) async {
     final resolver = buildStep.resolver;
+    final isLibrary = await resolver.isLibrary(buildStep.inputId);
+    final hasDartExtension = buildExtensions.containsKey(dartExtension);
 
-    if (!await resolver.isLibrary(buildStep.inputId) &&
-        buildExtensions.containsKey(dartExtension)) return;
+    if (!isLibrary && hasDartExtension) return;
 
     if (_generators.every((g) => g is GeneratorForAnnotation) &&
         !(await _hasAnyTopLevelAnnotations(
@@ -83,8 +84,10 @@ class _Builder extends Builder {
       return;
     }
 
-    final lib = await buildStep.resolver
-        .libraryFor(buildStep.inputId, allowSyntaxErrors: allowSyntaxErrors);
+    final lib = hasDartExtension
+        ? await buildStep.resolver
+            .libraryFor(buildStep.inputId, allowSyntaxErrors: allowSyntaxErrors)
+        : null;
     await _generateForLibrary(lib, buildStep);
   }
 
--- a/source_gen/lib/src/builder.dart
+++ b/source_gen/lib/src/builder.dart
@@ -38,6 +38,8 @@ class _Builder extends Builder {
   @override
   final Map<String, List<String>> buildExtensions;
 
+  static const dartExtension = '.dart';
+
   /// Wrap [_generators] to form a [Builder]-compatible API.
   _Builder(
     this._generators, {
@@ -48,7 +50,7 @@ class _Builder extends Builder {
     this.allowSyntaxErrors = false,
   })  : _generatedExtension = generatedExtension,
         buildExtensions = {
-          '.dart': [
+          dartExtension: [
             generatedExtension,
             ...additionalOutputExtensions,
           ]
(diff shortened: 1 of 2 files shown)


pratikpparikh:
@googlebot I signed it!

jakemac53:
This library is really not intended for use on non-Dart files - see the core [Builder](https://pub.dev/documentation/build/latest/build/Builder-class.html) class from the build system which is probably what you want.

I think we likely want to make the `library` argument non-nullable here as well when we migrate for null safety, which would be incompatible with this.
I don't think we want to support nullable

pratikpparikh:
@jakemac53 can you please explain the correct use case for LibraryBuilder as it stands today?  It seems to be useless, not a single example of LibraryBuilder.  Why limit to Dart files?  It be nice if you can explain the need for limitation of such nicely defined abstraction?

jakemac53:
LibraryBuilder, PartBuilder, and SharedPartBuilder all were built specifically for the case of consuming some Dart code and outputting more Dart code. This is baked into the apis at several levels.

For instance, `LibraryBuilder` only supports specifying the output extensions, not the input extensions. These are hard coded to be `.dart`. All of these classes also take `Generator`s as arguments, which are the things that actually do the codegen. These all are given a `LibraryReader` as the first argument, something which only makes sense if the input extension is a Dart file.

The majority of the functionality the package provides is based around _consuming_ dart code - it doesn't really do anything for you as far as emitting it (other than the shared part stuff).

It might be helpful to understand more what it is you want to use this package for, as opposed to the vanilla `build` package and its apis? What is this package providing that you see as missing from package:build?

pratikpparikh:
@jakemac53 thank you for your reply.  For my use case I have some generator that use PartBuilder and SharedPartBuilder which take benefit of annotation processing.  Using the same codebase I want create a version resilient code generator that will read json specification versions as input and output dart code library for the set of specification and its version.

i.e. we have a domain called ITSM and it has v1, v2, and v3.  The generator will generate a dart library that will be version resilient.  So if i use the the builder then i can't elegantly reuse the code write for other generator.

jakemac53:
> So if i use the the builder then i can't elegantly reuse the code write for other generator.

You should be able to re-use code here by creating shared functions here? Ultimately they both should just be building up output String objects, there shouldn't be any incompatible concepts between the two?

pratikpparikh:
@jakemac53, yes should be able to re-use code, but is it only me or does it not make sense to not block non dart input?

jakemac53:
The question is really "why does it make sense to allow non-Dart input"? I don't see a compelling case for that today, and I do see some arguments against it.

We don't want to broaden the scope of the package unnecessarily, narrowly targeted packages are easier to maintain. We wouldn't want to be hampered from making some design decision we would like to later on because of this choice.

I will give a very concrete example. Soon we will want to migrate this package to null safety. Today, the types on the [generate](https://pub.dev/documentation/source_gen/latest/source_gen/Generator/generate.html)  would all be able to stay unchanged - you are always given a non-null library reader and build step. With the change you propose here, the library reader would have to change to be `LibraryReader?` since it is nullable. That would be breaking for all existing generators, and it would mean in the future all existing generators have to deal with that nullability, which is not something we want.

So for that reason alone, I would need a very strong counter argument to allow non-dart files here.

jakemac53:
I am going to go ahead and close this pr - that doesn't mean the discussion has to end but I am pretty convinced that we shouldn't allow this for the reasons listed above.

pratikpparikh:
@jakemac53 cuz there are real use cases for non-Dart inputs*.  You are not broadening the scope of the package, you are making sure that the package is more widely usable. What design decision are going to be hampered?  if the package handles it correctly then nullsafety can be achieved correctly. I agree LibraryReader could be null and the package could easily handle that appropriately in _generateForLibrary.  Look it seems you have made up your mind and you want to limit the scope and you have the right to do that.  But there are very real use cases that would require non-dart inputs.  You can search the gitter and you will find people who have mentioned them, in any case I am not going to discuss this.  I will maintain my fork until I find a better abstraction that handles more comprehensive use cases and move on.

natebosch:
> But there are very real use cases that would require non-dart inputs.

Those use cases can use `package:build` APIs without depending on `package:source_gen`.

I don't understand how using `LibraryBuilder` with non-Dart input file is bringing you value. The only thing I can think of is that you get `dartfmt` run over the output, but you can get that in your own builder with less effort than trying to reuse `LibraryBuilder`.


The most relevant use case I can see for using `source_gen` at all with non-Dart inputs is to write Dart output which is included in the `.g.dart` file that other `SharedPartBuilder`s would use. You can accomplish that without changes in this package by using the same `appliesBuilder` configuration as you would if you were using a `SharedPartBuilder` and ensure your output is going to the right place. For instance if you have two source files `lib/foo.dart` and `lib/foo.json` and you want to output some dart code into `lib/foo.g.dart` based on metadata in `lib/foo.json` you should be able to have

```yaml
builders:
  my_builder:
    target: "my_builder"
    import: "package:my_builder/builder.dart"
    builder_factories: ["myBuilder"]
    build_extensions: {".json": [".my_output.g.part"]}
    auto_apply: dependents
    build_to: cache
    applies_builders: ["source_gen|combining_builder"]
```

PR #503, opened 2021-02-08
Use aliasElement over element for DartType

natebosch:


natebosch:
@scheglov see the second commit here - I had to add back a check for `GenericFunctionTypeElement` to get the tests to pass. Does this look correct to you?

scheglov:
No, this does not look right - we should be able to use just `aliasElement`.

natebosch:
> No, this does not look right - we should be able to use just `aliasElement`.

Without the extra code:

```
01:11 +167 -2: test/utils_test.dart: should return the name of a function type [E]                                                                                                                     
  UnimplementedError: (FunctionTypeImpl) dynamic Function()
  package:source_gen/src/utils.dart 34:3  typeNameOf
  test/utils_test.dart 41:12              main.<fn>
```

The test in question uses the legacy style typedef.

https://github.com/dart-lang/source_gen/blob/bacd2edcd710149e2282d86ed65594af9f39b4a3/source_gen/test/utils_test.dart#L25

scheglov:
Yes, I see that it does not work.
Support for `aliasElement` was added only recently, and the only version that supports it is `analyzer 0.42.nullsafety.0`.

scheglov (approved):

jakemac53 (approved):

PR #506, opened 2021-02-17
Allow latest pkg:analyzer

kevmoo:


source_gen/lib/src/library.dart
```diff
@@ -38,6 +38,7 @@ class LibraryReader {
     for (var cu in element.units) {
       yield* cu.accessors;
       yield* cu.enums;
+      // ignore: deprecated_member_use
```
jakemac53, line 41: Can we remove these and use the non-deprecated api? Or file an issue to remove them if we don't want to force the newest analyzer?

natebosch: https://github.com/dart-lang/source_gen/pull/503

I don't know what needs to happen in terms of landing and getting these published though. We may want to get a publish in with the wide lower bound before landing #503

kevmoo: That's my thought. This works fine. When we want to go w/ v2 of analyzer, this will go away and we'll have to deal.

kevmoo:
Can haz approve?

source_gen/lib/src/library.dart
jakemac53: My concern is that we don't actually test with old analyzers. How will we remember to remove these?

jakemac53: Oh I guess it would actually just break with the _new_ analyzer, which we should be testing with, so we should catch it.

jakemac53 (approved):

PR #513, opened 2021-03-09
Pass URI to AssetId.resolve

natebosch:
Updates for the breaking change in `package:build`.

- Use `Uri.parse` to convert to a URI, and migrate another call to the
  unnamed constructor since the URI is hardcoded and it can be resolved
  by hand.
- Override to the unpublished version of `build`.

### DIFF
--- a/source_gen/pubspec.yaml
+++ b/source_gen/pubspec.yaml
@@ -5,7 +5,7 @@ description: >-
 repository: https://github.com/dart-lang/source_gen
 
 environment:
-  sdk: ">=2.10.0 <3.0.0"
+  sdk: ">=2.11.999 <3.0.0"
 
 dependencies:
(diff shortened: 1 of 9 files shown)


source_gen/pubspec.yaml
```diff
@@ -5,12 +5,12 @@ description: >-
...
 environment:
-  sdk: ">=2.10.0 <3.0.0"
+  sdk: ">=2.11.999 <3.0.0"
 
 dependencies:
   analyzer: '>=0.42.0-nullsafety <2.0.0'
   async: ^2.0.7
-  build: ^1.4.0
+  build: ^2.0.0
   dart_style: ^1.0.0
```
kevmoo, line 14: All we need is `dart-style` updated and we could go full null-safe.

jakemac53 (approved):

Issue #493, opened 2020-11-22
Add support for nullsafety


vanlooverenkoen:
I am upgrading my package (Kiwi) https://github.com/vanlooverenkoen/kiwi/issues/51

But it seems that source_gen is not yet migrated to support nullsafety. Is there a timeline on when this could be added?

jakemac53:
This will be blocked on package:build, which is blocked on package:analyzer

vanlooverenkoen:
Blocked because of: https://github.com/dart-lang/build/issues/2920

kevmoo:
I believe this has been fixed!

kevmoo:
@vanlooverenkoen – do you want (1) the code in source_gen to be null safe? Or (2) just to support generating null-safe code?

We have (2) done, not (1).

vanlooverenkoen:
I want the code in code_gen to be nullsafe

kevmoo:
We need to get all of the downstream packages migrated, first.

You can still USE source_gen to generate null-safe code.

`json_serializable` does this now!

simolus3:
After 2.0 of `build` is released, we'd still be blocked by `dart_style` (and `build_verify` not supporting build 2.x, but that's not relevant now). 

Apart from that, I'm happy to migrate this package as well if no-one is working on it yet.

PR #520, opened 2021-03-16
Migrate to null-safety

kevmoo:
Closes https://github.com/dart-lang/source_gen/issues/493

### DIFF
--- a/source_gen/lib/src/output_helpers.dart
+++ b/source_gen/lib/src/output_helpers.dart
@@ -7,7 +7,7 @@ import 'package:async/async.dart';
 /// Converts [Future], [Iterable], and [Stream] implementations
 /// containing [String] to a single [Stream] while ensuring all thrown
 /// exceptions are forwarded through the return value.
-Stream<String> normalizeGeneratorOutput(Object value) {
+Stream<String> normalizeGeneratorOutput(Object? value) {
   if (value == null) {
     return const Stream.empty();
   } else if (value is Future) {
@@ -17,7 +17,7 @@ Stream<String> normalizeGeneratorOutput(Object value) {
   }
 
   if (value is Iterable) {
-    value = Stream.fromIterable(value as Iterable);
+    value = Stream.fromIterable(value);
   }
 
   if (value is Stream) {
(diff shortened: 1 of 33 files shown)


source_gen/lib/src/output_helpers.dart
```diff
@@ -26,7 +26,7 @@ Stream<String> normalizeGeneratorOutput(Object value) {
         return e.trim();
       }
 
-      throw _argError(e);
+      throw _argError(e as Object);
```
jakemac53, line 29: I would just do a `whereType<Object>()` above instead of the explicit null check and cast.

source_gen/pubspec.yaml
```diff
@@ -1,18 +1,18 @@
...
   Source code generation builders and utilities for the Dart build system
 repository: https://github.com/dart-lang/source_gen
 
 environment:
-  sdk: ">=2.11.999 <3.0.0"
+  sdk: ">=2.12.0 <3.0.0"
 
 dependencies:
-  analyzer: '>=0.42.0 <2.0.0'
+  analyzer: ^1.0.0
```
jakemac53, line 11: Do you know if you are relying on any of the changes to make some types be non-nullable now? (do the tests pass if you pin to this exactly?)

source_gen/lib/src/output_helpers.dart
kevmoo: `whereType` doesn't exist on Stream! That's what I did first!

CC @lrhn

source_gen/pubspec.yaml
kevmoo: will do a sanity run locally @ v1.0.0

kevmoo: dart_style requires ^1.10 so I bumped to that.

Tested locally. All is good!

source_gen/lib/src/output_helpers.dart
jakemac53: Oh it does in `stream_transform` I think, if we don't already depend on it though then this is fine.

jakemac53 (approved):

lazylazyllama:
@kevmoo When can we expect a stable release with this change?

kevmoo:
We're running some internal tests first. Maybe my end-of-day if we're lucky!

PR #521, opened 2021-03-17
Add changelog entries for removed APIs

natebosch:


natebosch:
@kevmoo @jakemac53 - FYI this is the type of info that would have made me do a TAP global instead of breaking internal users...

kevmoo:
Oh! I'm not done either. I'm doing some other cleanup. Good. They were waiting for a tap to drop..

kevmoo (approved):
My bad here!

Issue #490, opened 2020-11-13
Clarity with warning message


CharlesCarley:
It might be a little obvious after the fact, but as a newcomer to dart, I think that the warning message found in [builder.dart](https://github.com/dart-lang/source_gen/blob/master/source_gen/lib/src/builder.dart#L125) could be expanded.

If it could be changed from: 
```
    log.warning('Missing "part \'$part\';".');
```
to:
```
    log.warning('The include statement "part \'$part\';" was not found.');
```

It would provide a better sense of what is going wrong.

agent3bood:
The warning message was not clear at all, I has to google it to figure out what it meant.

natebosch:
Fixed with #528 . Thanks @agent3bood !

PR #528, opened 2021-04-19
Clear warning message

agent3bood:
Related to https://github.com/dart-lang/source_gen/issues/490


natebosch:
Thanks for the PR! I agree the old phrasing wasn't very clear. I think we can make it even more explicit than your suggestion so I push a commit that expands it further and pushes the directive to it's own line so it stands out and can be easier copied.

Issue #534, opened 2021-06-09
TypeChecker.fromRuntimeType doesn't work for types defined outside of `lib`


jakemac53:
The runtime type gets an absolute `file:` uri while the analyzer type gets an `asset:` uri. The `file:` uris should probably be normalized into `asset:` uris similar to how other types of uris are normalized.

I have a basic prototype of a fix that I will try to clean up and send a PR.

PR #535, opened 2021-06-09
fix TypeChecker.runtimeType for types that come from non-package: uris

jakemac53:
Fixes https://github.com/dart-lang/source_gen/issues/534

### DIFF
--- a/source_gen/lib/src/utils.dart
+++ b/source_gen/lib/src/utils.dart
@@ -149,5 +149,5 @@ final String rootPackageName = () {
         'Your pubspec.yaml file is missing a `name` field or it isn\'t '
         'a String.');
   }
-  return name as String;
+  return name;
 }();
--- a/source_gen/lib/src/utils.dart
+++ b/source_gen/lib/src/utils.dart
@@ -2,10 +2,13 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
+
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:build/build.dart';
 import 'package:path/path.dart' as p;
+import 'package:yaml/yaml.dart';
 
 /// Returns a non-null name for the provided [type].
 ///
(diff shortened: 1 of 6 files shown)


source_gen/lib/src/utils.dart
```diff
@@ -2,10 +2,13 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
```
kevmoo, line 5: Am I blind? Where is `dart:io` used here?

kevmoo (approved):

source_gen/lib/src/utils.dart
jakemac53: Its used in the initialization closure for the final variable `rootPackageName` :)

source_gen/test/type_checker_test.dart
```diff
@@ -173,37 +183,56 @@ void main() {
...
-            'package:source_gen/src/generator_for_annotation.dart#GeneratorForAnnotation'));
+      checkIterable: () => const TypeChecker.fromUrl('dart:core#Iterable'),
+      checkMap: () => const TypeChecker.fromUrl('dart:core#Map'),
+      checkHashMap: () => const TypeChecker.fromUrl('dart:collection#HashMap'),
+      checkGenerator: () => const TypeChecker.fromUrl(
+          'package:source_gen/src/generator.dart#Generator'),
+      checkGeneratorForAnnotation: () => const TypeChecker.fromUrl(
+          'package:source_gen/src/generator_for_annotation.dart#GeneratorForAnnotation'),
+      checkNonPublic: () => const TypeChecker.fromUrl(
+          'asset:source_gen/test/type_checker_test.dart#NonPublic'),
```
natebosch, line 234: This shows up differently internally so this test fails - internally we get a `google3://` not a `file://` URI.

If necessary we could patch a skip in or something...

source_gen/lib/src/utils.dart
```diff
@@ -94,6 +99,12 @@ Uri normalizeDartUrl(Uri url) => url.pathSegments.isNotEmpty
     ? url.replace(pathSegments: url.pathSegments.take(1))
     : url;
 
+Uri fileToAssetUrl(Uri url) {
+  if (!p.isWithin(p.current, url.path)) return url;
```
natebosch, line 103: Perhaps we could do something with the full package config to generalize this a bit more? For instance we could iterate over the package roots instead of assuming that `p.current` is within any package.

PR #538, opened 2021-06-28
ignore:unnecessary_non_null_assertion for future non-nullable LibraryElement.name

scheglov:
To re-apply https://dart-review.googlesource.com/c/sdk/+/205140

### DIFF
--- a/source_gen/lib/src/utils.dart
+++ b/source_gen/lib/src/utils.dart
@@ -41,7 +41,10 @@ String typeNameOf(DartType type) {
 ///
 /// Returns `null` if [element] is missing identifier.
 String nameOfPartial(LibraryElement element, AssetId source) {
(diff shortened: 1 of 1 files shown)


kevmoo (approved):

natebosch:
LGTM.

@jakemac53 - FYI the CFE now warns about this because it doesn't use the analyzers `// ignore` approach. If you try to run a script that imports `source_gen` directly, or snapshot it, you'd see.

```
lib/src/utils.dart:44:39: Warning: Operand of null-aware operation '!' has type 'String' which excludes null.
  if (element.name != null && element.name!.isNotEmpty) {
                                      ^
lib/src/utils.dart:45:20: Warning: Operand of null-aware operation '!' has type 'String' which excludes null.
    return element.name!;
                   ^
```

I verified that we don't surface this with `pub run build_runner build` though, so I think in the majority case it should be fine.

Issue #551, opened 2021-08-09
Apply dart_style's non-whitespace style fixes, like `dart format . --fix`


bradyt:
I am wondering if source_gen can implement something so that libraries, especially built_value, produce code that `dart format . --fix` won't modify.

The rest is copied from original issue at https://github.com/google/built_value.dart/issues/1086.

> I have my `*.g.dart` files committed in git, and I have `dart format . --fix` running in CI. I would like it if the code generated by built_value followed the advice found in dart format's --fix rules.
> 
> ```
> > dart help format
> ...
>     --fix                      Apply all style fixes.
> ```
> 
> - https://github.com/dart-lang/dart_style#style-fixes
> - https://github.com/dart-lang/dart_style/blob/2.0.3/lib/src/style_fix.dart
> 
> 
> To see what this looks like, try a fresh clone of built_value and try `dart format . --fix` there.
> 
> For example,
> 
> ```diff
> modified   example/lib/values.g.dart
> @@ -6,16 +6,16 @@ part of values;
>  // BuiltValueGenerator
>  // **************************************************************************
>  
> -Serializer<SimpleValue> _$simpleValueSerializer = new _$SimpleValueSerializer();
> +Serializer<SimpleValue> _$simpleValueSerializer = _$SimpleValueSerializer();
> ```
> 
> I didn't attempt to dig deep into the issue, I don't know if a related issue can be filed with source_gen, for example.
> 
> I'll try the following as a quick fix:
> 
> ```
> dart format $(find . \( -name '*.dart' ! -name '*.g.dart' \)) --fix
> ```

jakemac53:
Hmm we do run dart format generally on code (I think it allows you to override this also, but the default is to use dart format).

If there is an imperative api for applying fixes, I could see source_gen just doing that? At least by default.

bradyt:
Never mind, deleting comment (and this one) and I'll probably reply at other thread.

PR #552, opened 2021-08-09
auto-apply dartfmt fixes

jakemac53:
Closes https://github.com/dart-lang/source_gen/issues/551

I think this is pretty reasonable, especially since you can override the behavior.

@munificent are there any cases you know about where a fix would break code or all they all super safe?

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 1.2.2-dev
 
+* Enable the fix for single cascade statements when formatting Dart code.
+  * Individual builders can opt out of this by providing a custom `formatOutput`
+    function to the `PartBuilder` or `LibraryBuilder` constructors.
+
 ## 1.2.1
 
 - Allow reviving constants which are static fields defined on the class which
--- a/source_gen/lib/src/builder.dart
+++ b/source_gen/lib/src/builder.dart
@@ -391,7 +391,7 @@ Future<bool> _hasAnyTopLevelAnnotations(
   return false;
 }
 
-final _formatter = DartFormatter(fixes: StyleFix.all);
+final _formatter = DartFormatter(fixes: [StyleFix.singleCascadeStatements]);
 
 const defaultFileHeader = '// GENERATED CODE - DO NOT MODIFY BY HAND';
 
--- a/source_gen/lib/src/builder.dart
+++ b/source_gen/lib/src/builder.dart
@@ -391,7 +391,7 @@ Future<bool> _hasAnyTopLevelAnnotations(
   return false;
 }
 
-final _formatter = DartFormatter();
+final _formatter = DartFormatter(fixes: StyleFix.all);
 
 const defaultFileHeader = '// GENERATED CODE - DO NOT MODIFY BY HAND';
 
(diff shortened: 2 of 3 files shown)


natebosch:
My main concern is the potential performance impact of running fixes on all code when the vast majority won't need it. I'm slightly concerned about papering over stale behavior in generators, but I think probably the benefits outweigh that concern.

jakemac53:
Some of this is stale generators, but some of it is also just that its a lot easier to write a generator that way. For instance one of the fixes removes cascades where there is only a single one. But a generator might generate multiple cascades sometimes and not others and its pretty reasonable for it to just always generate some code using cascades, and then also pretty reasonable to have a tool come along and clean those up later?

jakemac53:
maybe @munificent could speak to any potential perf impacts, but we are already ok with the perf impacts of formatting generally here and non of these rules stand out to me as non-local (ie: slow) fixes.

jakemac53:
Fwiw this is the list of fixes:

```dart
  static const docComments = StyleFix._(
      'doc-comments', 'Use triple slash for documentation comments.');

  static const functionTypedefs = StyleFix._(
      'function-typedefs', 'Use new syntax for function type typedefs.');

  static const namedDefaultSeparator = StyleFix._('named-default-separator',
      'Use "=" as the separator before named parameter default values.');

  static const optionalConst = StyleFix._(
      'optional-const', 'Remove "const" keyword inside constant context.');

  static const optionalNew =
      StyleFix._('optional-new', 'Remove "new" keyword.');

  static const singleCascadeStatements = StyleFix._('single-cascade-statements',
      'Remove unnecessary single cascades from expression statements.');
```

natebosch (approved):

munificent:
I don't think there's anything wrong with this in general, but, for what it's worth, I have always designed the fixes with the idea in mind that a human would likely be around to vet the output. They don't promise to be as fire-and-forget as formatting changes are. It *should* be safe, but I can't promise that there won't be future fixes where you might not want to auto-apply them.

jakemac53:
> It _should_ be safe, but I can't promise that there won't be future fixes where you might not want to auto-apply them.

I think if that does come up, I would be willing to update this with a specific list instead of just applying all fixes.

Another alternative though could be to _only_ apply the cascade fix. All the other ones really should be fixed at the generator level, in an ideal world?

wdyt @natebosch ?

natebosch:
> Another alternative though could be to _only_ apply the cascade fix. All the other ones really should be fixed at the generator level, in an ideal world?

I have a slight preference for doing that, but I would be fine with either choice.

kevmoo:
I'm leaning against this. Maybe add it as something folks could opt-in to?

jakemac53:
> Maybe add it as something folks could opt-in to?

Folks could already opt into this today, but the _builder_ has to do it unfortunately.

They could add their own config to make it user configurable (a list of fixes to opt into, etc), but we don't have an easy way of exposing something uniform as far as user settings go really.

jakemac53:
> I'm leaning against this.

Are you opposed to doing even just the cascade fix? All the others are things that builder authors really could fix on their own I think.

natebosch:
@kevmoo - can you expand on your concerns?

kevmoo:
> @kevmoo - can you expand on your concerns?

Whitespace is a no-brainer. Other fixes seem like more of a stretch. I've had cases where autofixes mess up source code. 🤷 

It seems a bit too magic. I'mambivalentt, though. If ya'll think it's the right thing – run w/ it!

jakemac53:
@natebosch should I revive this? It only applies the cascade fix now.

natebosch:
Yeah, I think applying the cascade fix makes sense.

Issue #272, opened 2017-09-11
Generate code in another folder?


dam0vm3nt:
I think it could be interesting to group all the generated code in different folder from the original sources.
this makes simple to distinguish main source files from generated one (that one should not edit by hand).

Is it possible to add some options to `PartBuilder` and `LibraryBuilder` in order to specify a different folder (even restricted to `lib`) where to put generated files ?

jakemac53:
Seems reasonable, I think we probably would want to keep outputs in the same top level directory (lib/web/etc) as the original inputs, and then mirror the rest of the path under the generated files directory.

So if you had a file at `lib/src/models/model.dart` and set your output dir to `lib/src/generated`, we would output the part file at `lib/src/generated/src/models/model.g.dart`.

What do you think?

The main issue I see is that the part statements would be pretty long/ugly. However we need to mirror the directory structure to make sure we don't have conflicts....

dam0vm3nt:
LGTM

I don't think the part import path is actually a problem. To me it is preferably compared to having generated files mixed with original one.

natebosch:
This won't be easy but should be feasible. Do we need it to be configurable or should it be a boolean flag on `Builder` and we hardcode the directory name? Is it worth the extra cognitive load of having a new field on `Builder` - how much will it get used? How will the builder find the `AssetId`, will we need a new utility there like `AssetId.inGeneratedDirectory`?

Is this a property of the code generator, or of the usage? If you're using `built_value` should you get to choose to use the generated directory, or would you need to file an issue requested that it uses the generated directory?

jakemac53:
Ah  ya I guess this is more challenging than I realized at first because of the migration away from `declareOutputs` and to a direct file extension conversion... originally I wasn't thinking we would need any changes in package:build.

jakemac53:
Also fwiw we are moving more towards not outputting files in the main source directory any more, but instead outputting them to a hidden directory and then materializing a merged view in some output folder.

This is because we are adding support for generating files in other packages so we have to create our own view of the world to prevent ourselves from polluting the users pub cache.

Once we are fully migrated over to this model then that probably closes this issue as well?

natebosch:
I think there could still be a use case for generating files to a side directory and committing those files and uploading them to pub, though I'd also be happy to hear that just building to a cache solves the issue.

dam0vm3nt:
[...]

PR #554, opened 2021-08-17
Configurable outputs for the combining builder

simolus3:
This adds support for custom output locations, including outputs in different directories, to the combining builder.

- Make output extensions configurable through builder options. The options are validated to ensure the combining builder is running on Dart files and emits Dart files
- Move the check ensuring we have a `part` statement from the `SharedPartBuilder` to the `CombiningBuilder`. The former can't know the correct output location anymore.

This is related to #272, where the implementation in `build` requires cooperation from builders. Given that most generated files come from shared part generators in practice, this would enable generating to different folders for most users.

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.1.0-dev
+
+* Add the `build_extensions` option to `combining_builder`, allowing output
+  files to be generated into a different directory.
+
 ## 1.0.5
 
 * Fix a bug with reviving constant expressions which are fields defined on a
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.1.0-dev
+## 1.1.0
 
 * Add the `build_extensions` option to `combining_builder`, allowing output
   files to be generated into a different directory.
--- a/source_gen/README.md
+++ b/source_gen/README.md
@@ -117,6 +117,41 @@ targets:
 If you provide a builder that uses `SharedPartBuilder` and `combining_builder`,
 you should document this feature for your users.
 
+### Generating files in different directories
+
+When using shared-part builders which apply the `combining_builder` as part of
+the build, the output location for an input file can be changed.
+By default, a `.g.dart` file next to the input is generated.
+
+To change this, set the `build_extensions` option on the combining builder. In
+the options, `build_extensions` is a map from `String` to `String`, where the
+key is matches inputs and the value is a single build output.
+For more details on build extensions, see [the docs in the build package][outputs].
+
+For example, you can use these options to generate files under `lib/generated`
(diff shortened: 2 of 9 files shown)


source_gen/README.md
```diff
@@ -117,6 +117,33 @@ targets:
 If you provide a builder that uses `SharedPartBuilder` and `combining_builder`,
 you should document this feature for your users.
 
+### Generating files in different directories
+
+When using shared-part builders which apply the `combining_builder` as part of
+the build, the output location for an input file can be changed.
+By default, a `.g.dart` file next to the input is generated.
+
+To change this, set the `build_extensions` option on the combining builder. In
```
jakemac53, line 126: See https://github.com/dart-lang/build/issues/3172 - we should tell people to add a dev dependency on source_gen and require at least the version this feature ships in.

source_gen/lib/builder.dart
```diff
@@ -41,6 +46,38 @@ Builder combiningBuilder([BuilderOptions options = BuilderOptions.empty]) {
   return builder;
 }
 
+Map<String, List<String>> _validatedBuildExtensionsFrom(
+    Map<String, dynamic> optionsMap) {
+  final extensionsOption = optionsMap.remove('build_extensions');
+  if (extensionsOption is! Map) return _defaultExtensions;
```
jakemac53, line 52: This should throw an argument error if we get something non-null that isn't a map

source_gen/lib/builder.dart
```diff
@@ -41,6 +46,43 @@ Builder combiningBuilder([BuilderOptions options = BuilderOptions.empty]) {
...
+    if (output is! String || !output.endsWith('.dart')) {
+      throw ArgumentError('Invalid output extension `$output`. It should be a '
+          'string ending with `.dart`');
+    }
+
+    result[input] = [output];
+  }
+
+  if (result.isEmpty) {
+    log.warning('Using default extensions as all values from the '
```
jakemac53, line 78: I think this should probably throw as well - I think this is only reachable if somebody provides an empty config, which should probably just be an error.

jakemac53 (approved):

jakemac53:
I filed https://github.com/dart-lang/sdk/issues/46939 for the analyzer failure, we may need to mark that test as skipped for now

simolus3:
It works with the newest analyzer, I've fixed deprecation warnings which requires using the latest version as a lower bound.

I agree that we should eventually make the tests more reliable, I also had some of them failing when running in isolation because of the "Generating SDK summary" logs.

source_gen/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.1.0-dev
```
jakemac53, line 1: I would probably drop the `-dev` and we can go ahead and publish (pubspec also). I don't think we have any other pending changes in this package.

PR #556, opened 2021-09-27
Drop pkg:pedantic, require Dart 2.14, prepare for release

kevmoo:


source_gen/lib/src/builder.dart
```diff
@@ -162,7 +162,7 @@ source formatter.''',
       );
     }
 
-    unawaited(buildStep.writeAsString(outputId, genPartContent));
+    buildStep.writeAsString(outputId, genPartContent).ignore();
```
kevmoo, line 165: Or could we just `await` this? @jakemac53 @natebosch ?

natebosch: It wouldn't make a difference here. We were more focused on demonstrating the non-awaiting early on with `build_runner` because I think we expected more places where it would make a difference. In practice I think it rarely comes up that it makes a difference but @jakemac53 might know some places where it has mattered.

I would be fine adding an `await` here.

jakemac53: I would just add an `await`. We don't require you to await these but its just easier and its very unlikely to have a visible performance impact.

kevmoo:
PTAL!

natebosch (approved):

PR #564, opened 2021-11-15
Configurable outputs for the PartBuilder

sperochon:
This adds support for custom output locations, including outputs in different directories, to the PartBuilder.

- Make output extensions configurable through builder options. The options are validated to ensure the PartBuilder is running on Dart files and emits Dart files (re-use existing validatedBuildExtensionsFrom)

This was done previously for the combining builder (https://github.com/dart-lang/source_gen/pull/554). This PR is for PartBuilder.

I've added some tests. All tests are passing. But as I'm not an expert of source_gen code, could you please review carefully this PR, especially :

file: source_gen/lib/src/builder.dart
new line:
`final outputId = buildStep.allowedOutputs.single;`
replacing:
`final outputId = buildStep.inputId.changeExtension(_generatedExtension);`

### DIFF
--- a/source_gen/README.md
+++ b/source_gen/README.md
@@ -119,11 +119,13 @@ you should document this feature for your users.
 
 ### Generating files in different directories
 
-When using shared-part builders which apply the `combining_builder` as part of
-the build, the output location for an input file can be changed.
-By default, a `.g.dart` file next to the input is generated.
+The output location for an input file can be changed:
+- when using `PartBuilder`,
+- when using `SharedPartBuilder` which apply the `combining_builder` as part of
+the build,
 
-To change this, set the `build_extensions` option on the combining builder. In
+By default, a `.g.dart` or `.some_name.dart` file is generated next to the input.
+To change this, set the `build_extensions` option on the corresponding builder. In
 the options, `build_extensions` is a map from `String` to `String`, where the
 key is matches inputs and the value is a single build output.
 For more details on build extensions, see [the docs in the build package][outputs].
@@ -139,6 +141,11 @@ targets:
         options:
           build_extensions:
             '^lib/{{}}.dart': 'lib/generated/{{}}.g.dart'
+
+      some_cool_builder:
+        options:
+          build_extensions:
+            '^lib/models/{{}}.dart': 'lib/models/generated/{{}}.foo.dart'
 ```
 
 Remember to change the `part` statement in the input to refer to the correct
--- a/source_gen/README.md
+++ b/source_gen/README.md
@@ -137,11 +137,13 @@ with the following build configuration:
 targets:
   $default:
     builders:
+      # A SharedPartBuilder which uses the combining builder
       source_gen|combining_builder:
         options:
           build_extensions:
             '^lib/{{}}.dart': 'lib/generated/{{}}.g.dart'
 
+      # A PartBuilder or LibraryBuilder
       some_cool_builder:
         options:
           build_extensions:
--- a/source_gen/README.md
+++ b/source_gen/README.md
@@ -121,8 +121,8 @@ you should document this feature for your users.
 
 The output location for an input file can be changed:
 - when using `PartBuilder` or `LibraryBuilder`.
-- when using `SharedPartBuilder` which apply the `combining_builder` as part of
-the build,
+- when using `SharedPartBuilder` which apply the `combining_builder` as
+part of the build.
 
 By default, a `.g.dart` or `.some_name.dart` file is generated next to the input.
 To change this, set the `build_extensions` option on the corresponding builder. In
--- a/source_gen/README.md
+++ b/source_gen/README.md
@@ -120,7 +120,7 @@ you should document this feature for your users.
 ### Generating files in different directories
 
 The output location for an input file can be changed:
-- when using `PartBuilder`,
+- when using `PartBuilder` or `LibraryBuilder`.
 - when using `SharedPartBuilder` which apply the `combining_builder` as part of
 the build,
 
--- a/source_gen/lib/src/builder.dart
+++ b/source_gen/lib/src/builder.dart
@@ -48,9 +48,10 @@ class _Builder extends Builder {
     List<String> additionalOutputExtensions = const [],
     String? header,
     this.allowSyntaxErrors = false,
-    BuilderOptions options = BuilderOptions.empty,
+    BuilderOptions? options,
   })  : _generatedExtension = generatedExtension,
-        buildExtensions = validatedBuildExtensionsFrom(Map.of(options.config), {
+        buildExtensions = validatedBuildExtensionsFrom(
+            options != null ? Map.of(options.config) : null, {
           '.dart': [
             generatedExtension,
             ...additionalOutputExtensions,
@@ -66,8 +67,7 @@ class _Builder extends Builder {
       throw ArgumentError(
           'A standalone file can only be generated from a single Generator.');
     }
-    if (options != BuilderOptions.empty &&
-        additionalOutputExtensions.isNotEmpty) {
+    if (options != null && additionalOutputExtensions.isNotEmpty) {
       throw ArgumentError(
           'Either `options` or `additionalOutputExtensions` parameter '
           'can be given. Not both.');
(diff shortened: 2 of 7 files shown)


source_gen/lib/src/builder.dart
```diff
@@ -94,8 +102,7 @@ class _Builder extends Builder {
     // library/part definitions because users expect some files to be skipped
     // therefore they do not have "library".
     if (generatedOutputs.isEmpty) return;
-    final outputId = buildStep.inputId.changeExtension(_generatedExtension);
-
+    final outputId = buildStep.allowedOutputs.single;
```
jakemac53, line 105: There might be more than one if there are `additionalOutputExtensions`. Presumably the ordering is kept here though so you could just grab the first?

```suggestion
    final outputId = buildStep.allowedOutputs.first;
```

If this didn't break tests we must not be testing `additionalOutputExtensions`?

source_gen/test/builder_test.dart
```diff
@@ -228,69 +228,120 @@ void main() {
...
-        generateFor: {
-          '$_pkgName|lib/a.dart'
-        },
-        outputs: {
-          '$_pkgName|lib/a.foo.dart':
-              decodedMatches(startsWith('$_customHeader\n\npart of')),
-        });
-  });
+  group('PartBuilder', () {
+    test('PartBuilder uses a custom header when provided', () async {
```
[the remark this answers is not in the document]
sperochon: Aaaaaahhh. I did what you suggest at first, but it appears that when running individual test with same name as other in another group, both are executed...

For example: running individual test of group 'PartBuilder' called 'includes matching language version in all parts' will also launch individual test 'includes matching language version in all parts' of group 'CombiningBuilde
Capture d’écran 2021-11-15 à 17 11 32
r'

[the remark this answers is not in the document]
jakemac53: Oh weird, ya I think that is just a side effect of how the IDE integration works... but I wouldn't duplicate the names to resolve it.

We actually recently released some better support in the test packages for running tests by line number which should give the IDE tooling a way to fix this in the future.

source_gen/lib/src/builder.dart
```diff
@@ -35,23 +35,21 @@ class _Builder extends Builder {
   final bool allowSyntaxErrors;
 
   @override
-  final Map<String, List<String>> buildExtensions;
+  late final Map<String, List<String>> buildExtensions;
```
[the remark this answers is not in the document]
sperochon: Thanks for the tip!

source_gen/lib/src/builder.dart
```diff
@@ -62,6 +60,16 @@ class _Builder extends Builder {
       throw ArgumentError(
           'A standalone file can only be generated from a single Generator.');
     }
+    
+    final optionsMap = Map<String, dynamic>.from(options.config);
```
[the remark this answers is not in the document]
sperochon: Thanks, I now know the diff between Map.from and Map.of

source_gen/test/builder_test.dart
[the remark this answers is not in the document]
sperochon: Ok, fixed!

jakemac53:
Another question I have here is how this should work with `additionalOutputExtensions`? I think today it might essentially not be compatible, because we don't allow non-dart outputs.

I am not sure of the best way to reconcile these two things being passed together? We could just explicitly disallow it for now, or we could add the `additionalOutputExtensions` to the map returned from the `validatedBuildExtensionsFrom` function instead of passing them as part of the defaults?

sperochon:
Ok, I do it (didn't know if you would have preferred another PR...)

jakemac53:
Assuming its minimal, adding to this PR sounds good. If it is larger then another PR also works!

source_gen/test/builder_test.dart
```diff
@@ -344,6 +344,31 @@ part "a.foo.dart";'''
     });
   });
 
+  group('LibraryBuilder', () {
+    group('with build_extensions', () {
+      test('generates relative `path of` for output in different directory',
```
jakemac53, line 349: This name doesn't seem quite right, maybe:
```suggestion
      test('outputs to the correct location',
```

sperochon: whoops

jakemac53:
We do still need to figure out what to do with `additionalOutputExtensions` here, but otherwise its looking good.

source_gen/lib/src/builder.dart
```diff
@@ -66,6 +66,12 @@ class _Builder extends Builder {
       throw ArgumentError(
           'A standalone file can only be generated from a single Generator.');
     }
+    if (options != BuilderOptions.empty &&
```
jakemac53, line 69: I think `BuilderOptions.empty` is the default thing that will end up flowing through here if the builder is simply not configured.

Ultimately it might actually be best to just make the `BuilderOptions` argument be nullable instead of defaulting to empty. Then you can check for null here which should more accurately do what we want.

The `validatedBuildExtensionsFrom` function could also then accept a nullable BuilderOptions and bail early if it gets a null.

sperochon: > make the BuilderOptions argument be nullable instead of defaulting to empty
-> commited

> The `validatedBuildExtensionsFrom` function could also then accept a nullable BuilderOptions and bail early if it gets a null.

The thing is that the `validatedBuildExtensionsFrom` function is also called by the combiningBuilder function which already sends an `optionsMap` (and not `options`). So I can't change the argument to a nullable `BuilderOptions`. I've made it nullable though to manage cases where `options` is null.

source_gen/README.md
```diff
@@ -119,11 +119,13 @@ you should document this feature for your users.
 
 ### Generating files in different directories
 
-When using shared-part builders which apply the `combining_builder` as part of
-the build, the output location for an input file can be changed.
-By default, a `.g.dart` file next to the input is generated.
+The output location for an input file can be changed:
+- when using `PartBuilder`,
```
jakemac53, line 123:
```suggestion
- when using `PartBuilder` or `LibraryBuilder`.
```

source_gen/README.md
```diff
@@ -119,11 +119,13 @@ you should document this feature for your users.
...
+- when using `SharedPartBuilder` which apply the `combining_builder` as part of
+the build,
```
jakemac53, line 125:
```suggestion
- when using `SharedPartBuilder` which apply the `combining_builder` as
part of the build.
```

source_gen/README.md
```diff
@@ -139,6 +141,11 @@ targets:
         options:
           build_extensions:
             '^lib/{{}}.dart': 'lib/generated/{{}}.g.dart'
+
+      some_cool_builder:
```
jakemac53, line 145:
```suggestion
      # A PartBuilder or LibraryBuilder
      some_cool_builder:
```

I can't comment on the previous config but I would add a comment on that one also that its configuring a `SharedPartBuilder` which uses the combining builder

jakemac53 (commented):
Ok, I did one final pass here. Sorry for all the back and forth, but last one from me I promise! haha

jakemac53:
Also see the analyzer/dartformat failures on CI, should be quick to clean up.

source_gen/README.md
sperochon: ok, done

jakemac53 (approved):

sperochon:
> Ok, I did one final pass here. Sorry for all the back and forth, but last one from me I promise! haha

Thanks for all your advices! 👍😀

d-markey:
> We do still need to figure out what to do with `additionalOutputExtensions` here, but otherwise its looking good.

There still are a few problems with `validatedBuildExtensionsFrom` and `additionalOutputExtensions`/`options`. If a `PartBuilder` generates additional output files and the user wants to emit to another directory for instance, `additionalOutputExtensions` cannot be used because the `PartBuilder` later enforces only one of `options` or `additionalOutputExtensions`. So to take into account the user's config, `options` must be passed. But then `validatedBuildExtensionsFrom` rejects it if it contains several extensions (`List<String>`) instead of a `String`.

I have proposed PR https://github.com/dart-lang/source_gen/pull/647 to allow for configuration with multiple output files.

Note: if no config was provided by the user in `build.yaml`, the code that creates the `PartBuilder` should override `options` with the default `build_extensions` before calling `validatedBuildExtensionsFrom`. `additionalOutputExtensions` cannot be used to set the default values since it's rejected by `PartBuilder` when `options` is provided. Maybe it could be renamed to `defaultOutputExtensions` and `PartBuilder` should allow it together with `options`?

PR #571, opened 2021-11-24
Improve missing part handling on Windows

### LINKED ISSUE #570: Files only generated in cache(?), not in source tree
Hi all,

```
    • Flutter version 2.5.3 at C:\Maarten\local\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 18116933e7 (6 weeks ago), 2021-10-15 10:46:35 -0700
    • Engine revision d3ea636dc5
    • Dart version 2.14.4
```

I'm trying to follow the example from Simon Binder (https://simonbinder.eu/posts/build_directory_moves/) on how to place the json_serializable files into a separate directory. What I'm observing is that the generated files do get created, but they are only available under the ".dart_tools/build" directory, they do not show up in the main source tree as expected.

pubspec.yaml (only relevant bits):
```
dependencies:
  json_annotation: ^4.3.0

dev_dependencies:
  build_runner: ^2.1.5
  source_gen: ^1.1.1
  json_serializable: ^6.0.0
```

build.yaml:
```
targets:
  $default:
    builders:
      source_gen|combining_builder:
        options:
          build_extensions:
            '^lib/{{}}.dart': 'lib/generated/{{}}.g.dart'
```

All my "json classes" are under `<project root>/lib/api/model`. Expected result is that I see the generated files under `<project root>/lib/generated/api/model`. But they don't show up there. I only see them under `<project root>/.dart_tools/build/generated/<project name>/lib/generated/api/model`.

What am I doing wrong here? Based on other issues I found, I suspect this has something to do with the combining_builder?

simolus3:
The combining part builder checks for the existence of a `part` statement in the source file, ensuring that generated code is included properly.

To do this, it computes the relative path of the configured output and the input asset using the `path` library. However, it uses the platform-specific default path context, even though Dart imports always use an URL-based context. This is not a problem if the input and the output are in the same directory, or if the native system uses a forward slash as a path separator. When using custom build extensions putting generated files in a different directory, this turns out to be a problem on Windows.

This PR ensures a consistent checking mechanism by always using the URL context. Fixes #570.

I'm not sure how to test this properly. Ideally we should run the CI on Windows too I guess, but I haven't looked into how this could be done with `mono_repo`.

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -9,6 +9,7 @@
   constructor for this to work.
 - Fix a bug finding source locations for reporting unresolved annotations on
   parameters.
+- Fix a bug checking for `part` statements on Windows.
 
 ## 1.1.1
 
--- a/source_gen/lib/src/utils.dart
+++ b/source_gen/lib/src/utils.dart
@@ -8,7 +8,7 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:build/build.dart';
-import 'package:path/path.dart' as p show url;
+import 'package:path/path.dart' as p;
 import 'package:yaml/yaml.dart';
 
 /// Returns a non-null name for the provided [type].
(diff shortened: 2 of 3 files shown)


jakemac53:
Adding windows testing in https://github.com/dart-lang/source_gen/pull/572

jakemac53:
@simolus3 you should be able to merge with master now and then unskip all tests that are skipped due to this issue (search for `issues/570`).

natebosch (approved):
Thanks!

boekhold:
When could we reasonably expect an updated source_gen package on pub.dev that contains this fix?

jakemac53:
I can work on getting this released this week - we need to first roll it internally though and make sure we don't break anything there.

jakemac53:
@boekhold this is released now (version `1.2.0`)

PR #576, opened 2021-12-03
Add a note about not importable annotations

natebosch:
The build systems run on the Dart VM only, so there are limitations on
what imports can be used by builders. Add documentation describing how
to work around this limitation by overriding the `TypeChecker` with
something other than `fromRuntime`.

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 1.2.1-dev
+
 ## 1.2.0
 
 - Include the `LibraryElement` in `LibraryReader.allElements`,
--- a/source_gen/lib/src/generator_for_annotation.dart
+++ b/source_gen/lib/src/generator_for_annotation.dart
@@ -39,7 +39,14 @@ import 'type_checker.dart';
 /// extension, are not searched for annotations. To operate on, for instance,
 /// annotated fields of a class ensure that the class itself is annotated with
 /// [T] and use the [Element] to iterate over fields. The [TypeChecker] utility
-/// may be helpful to check which elements have a given annotation.
+/// may be helpful to check which elements have a given annotation if the
+/// generator should further filter it's target based on annotations.
+///
+/// If the annotation type cannot be imported on the Dart VM, for example if it
+/// imports `dart:html` or `dart:ui`, then the default behavior of using
+/// `TypeChecker.fromRuntime` is not feasible. In these cases extend
+/// `GeneratorForAnnotation<void>` and override the [typeChecker] member with a
+/// checker matching the annotation type.
 abstract class GeneratorForAnnotation<T> extends Generator {
   const GeneratorForAnnotation();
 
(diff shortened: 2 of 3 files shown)


kevmoo (approved):

jakemac53 (approved):
Should we just provide a `GeneratorForMatchingAnnotation` which takes a TypeChecker?

kevmoo:
> Should we just provide a `GeneratorForMatchingAnnotation` which takes a TypeChecker?

Just as a factory even, would be great. A follow-up is good here.

natebosch:
> > Should we just provide a `GeneratorForMatchingAnnotation` which takes a TypeChecker?
> 
> Just as a factory even, would be great. A follow-up is good here.

The existing design imposes an `extends` already - you are intended to override `generateForAnnotatedElement`. 

It would be nice to change that and allow composition but 🤷‍♂️ 

Given the current design, I think `@override` fits. WDYT?

natebosch:
> which class you extend would change (and what you pass to the super constructor)

Adding another class to the public API would increase the surface area of the package and IMO make things harder to find.

jakemac53:
> Adding another class to the public API would increase the surface area of the package and IMO make things harder to find.

🤷‍♂️ I don't really agree, I think its significantly better than doing the whole `GeneratorForAnnotation<void>` thing and then overriding the type checker. Overriding the type checker in that way is definitely a bit funky, and you are relying essentially on how that thing is being used internally in the super class...

natebosch:
I'm not able to find a design I like here with another class (especially because I can't think of a nice name).

With a breaking change I could imagine something different, maybe where we drop the generic and you need a constructor when forwards to `super.forType(someType)` or `super.forTypeChecker(someTypeChecker)`. If we're going with a breaking change we may as well drop the need for `extends` entirely though...

kevmoo:
@natebosch – thoughts on this?

kevmoo:
Uh...this is old @natebosch

natebosch:
@jakemac53 - do you think this is worth landing as is?

source_gen/CHANGELOG.md
```diff
@@ -39,6 +42,7 @@
   represents their type. Previously we checked this pattern only for enums,
   however there are enum-like usages in classes which are not enums.
 - Allow the latest version of `package:analyzer`.
+>>>>>>> master
```
jakemac53, line 45:
```suggestion
```

jakemac53:
I still really don't like the `GeneratorForAnnotation<void>` suggestion personally. Should we instead just tell people to split up their annotation libraries so they don't have those imports?

natebosch:
> Should we instead just tell people to split up their annotation libraries so they don't have those imports?

I don't know if that is always feasible - some annotations could use `dart:ui` classes as arguments.

natebosch:
@jakemac53 - is this implementation of `GeneratorForMatchingAnnotation` match what you were thinking?

jakemac53 (approved):

Issue #577, opened 2021-12-09
Cannot revive ProcessStartMode.normal


srawlins:
In mockito, we might try to revive a constant, `ProcessStartMode.normal` ([docs](https://api.dart.dev/stable/2.15.0/dart-io/ProcessStartMode/normal-constant.html)), but source_gen always gives back `ProcessStartMode._internal()` (the constructor used by `normal`).

I can reproduce this in constants_test, where the LibraryElement for dart.io has _zero_ types, classes, etc. and so never tries `ProcessMode.normal` as a Revive. Very strange... :(

Similar to https://github.com/dart-lang/source_gen/issues/547, but not the same bug, I believe.

PR #578, opened 2021-12-09
reviving: Prefer static constant fields on constant object's class

srawlins:
Fixes https://github.com/dart-lang/source_gen/issues/577

This should also be a faster heuristic; when looking for an object of type Foo, check Foo's static fields first. This was already done for enums; I see no reason to do it for all classes.

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.2.1-dev
+
+- Allow reviving constants which are static fields defined on the class which
+  represents their type. Previously we checked this pattern only for enums,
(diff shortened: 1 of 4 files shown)


srawlins:
CC @natebosch if you have time to review this week.

kevmoo:
What's the plan here? @natebosch ?

natebosch (approved):

kevmoo:
Should we have a changelog entry here, @srawlins ?

kevmoo:
Ha! @natebosch reads my mind

srawlins:
Thanks!

PR #591, opened 2022-03-04
Use 'CompilationUnitElement.classes' instead of 'types'.

scheglov:
We would like to remove it.
[https://dart-review.googlesource.com/c/sdk/+/235290](https://www.google.com/url?sa=D&q=https%3A%2F%2Fdart-review.googlesource.com%2Fc%2Fsdk%2F%2B%2F235290)


kevmoo (approved):
Any need to bump the min pkg:analyzer constraint?

scheglov:
Oh, there is already https://github.com/dart-lang/source_gen/pull/590

PR #600, opened 2022-05-05
Fix a misleading regex

### LINKED ISSUE #599: Incorrect (too permissive) regex in CombiningBuilder?
The `restrictedPatter` seems to handle a literal `.` (dot) incorrectly, it should escape it. The way it is now it means the regex allows any character in that place, including a dot: https://github.com/dart-lang/source_gen/blob/140828cf2bf6747394ad19029cde787c615029d3/source_gen/lib/builder.dart#L90

natebosch:
Closes #599

This looks like a bug, but it can't currently surface as a problem
because the character at that location is already guaranteed to be
correct due to the glob pattern.

We cannot shorten the regex to check only the end of the file name
because we need to also disallow names which have extra `.` where there
should be a single part identifier. We could simplify the regex to just
check the right number of characters up to the `.` and count on the glob
for the earlier part, but I think it is more readable with the full
regex.

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 1.2.3-dev
+
 ## 1.2.2
 
 * Enable the fix for single cascade statements when formatting Dart code.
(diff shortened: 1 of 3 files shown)


jakemac53 (approved):

jakemac53:
I believe this is causing https://github.com/dart-lang/source_gen/issues/603.

Issue #605, opened 2022-06-03
Found 2 files excluded from sound null safety


devoncarew:
Found 2 files excluded from sound null safety:

```
source_gen/test/builder_test.dart:517:// @dart=2.9
source_gen/test/builder_test.dart:528:// @dart=2.9
```

See https://github.com/dart-lang/sdk/issues/49169 for more context.

PR #606, opened 2022-06-03
Bump min SDK used in language version test to 2.12

kevmoo:
Fixes https://github.com/dart-lang/source_gen/issues/605

### DIFF
--- a/source_gen/test/builder_test.dart
+++ b/source_gen/test/builder_test.dart
@@ -296,13 +296,13 @@ void main() {
           header: '',
         ),
         {
-          '$_pkgName|lib/a.dart': '''// @dart=2.9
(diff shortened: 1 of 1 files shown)


source_gen/test/builder_test.dart
```diff
@@ -296,13 +296,13 @@ void main() {
           header: '',
         ),
         {
-          '$_pkgName|lib/a.dart': '''// @dart=2.9
+          '$_pkgName|lib/a.dart': '''// @dart=2.12
```
kevmoo, line 299: This is only in generated test code.

Bumping to 2.12 to avoid the falso positive.

devoncarew (approved):

PR #608, opened 2022-07-07
Add 'part of' to parts, otherwise it will not be a part.

scheglov:


source_gen/test/builder_test.dart
```diff
@@ -93,18 +93,6 @@ void main() {
     );
   });
 
-  test('Allow no "library"  by default', () async {
```
jakemac53, line 96: why is this one removed?

jakemac53: ah ok I see

jakemac53 (approved):

PR #611, opened 2022-07-25
Ignore deprecated api usage from pkg:analyzer

kevmoo:


source_gen/lib/src/constants/revive.dart
```diff
@@ -40,6 +40,7 @@ Revivable reviveInstance(DartObject object, [LibraryElement? origin]) {
   if (element is MethodElement && element.isStatic) {
     return Revivable._(
       source: url.removeFragment(),
+      // ignore: deprecated_member_use
```
jakemac53, line 43: can we not fix the use of the deprecated member?

kevmoo: Sure. But we'll still have to bump the analyzer version for the next major release.

Figured it'd be better to just do the tiny thing now and roll everything later. 🤷

jakemac53: I would just fix it now myself and up the analyzer constraint?

kevmoo: Ugh. Fine.

kevmoo:
@jakemac53 – PTAL

jakemac53 (approved):

PR #613, opened 2022-07-28
Move lints to use package:lints

kevmoo:
Migrate analyzer bits from strong-mode to strict- bits – and fix

Remove duplicative entries


jakemac53 (approved):

source_gen/test/output_helpers_test.dart
```diff
@@ -49,7 +49,7 @@ void main() {
 
     _testFunction(
       'new Future.error',
-      () => Future.error(ArgumentError('Error message')),
+      () => Future<void>.error(ArgumentError('Error message')),
```
natebosch, line 52: [nit] Unless this causes a static error...

```suggestion
      () => Future<Never>.error(ArgumentError('Error message')),
```

natebosch (approved):

PR #614, opened 2022-08-07
Store Annotated Elements in GeneratorForAnnotation

lyledean1:
Store annotated elements inside GeneratorForAnnotation so you can compare elements to see if its the last element present 

Example would be generating an Enum based off all annotated elements of type T i.e returning the completed Enum string on the last element

``` dart
class EnumGenerator extends GeneratorForAnnotation<Example> {
      @override
       Future<String> generateForAnnotatedElement(
          Element element,
          ConstantReader annotation,
          BuildStep buildStep) async {
            if (annotatedElements!.last.element == element) {
               // do something for the last element 
             } 
      }
    }
```

### DIFF
--- a/source_gen/lib/src/generator_for_annotation.dart
+++ b/source_gen/lib/src/generator_for_annotation.dart
@@ -41,15 +41,17 @@ import 'type_checker.dart';
 /// [T] and use the [Element] to iterate over fields. The [TypeChecker] utility
 /// may be helpful to check which elements have a given annotation.
 abstract class GeneratorForAnnotation<T> extends Generator {
-  const GeneratorForAnnotation();
+  GeneratorForAnnotation();
+  Iterable<AnnotatedElement>? annotatedElements;
 
   TypeChecker get typeChecker => TypeChecker.fromRuntime(T);
 
   @override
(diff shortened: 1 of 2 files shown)


natebosch:
Thanks for the suggestion!

As written this is not more efficient than implementing your builder as

``` dart
class EnumGenerator extends GeneratorForAnnotation<Example> {
  @override
  Future<String> generateForAnnotatedElement(
      Element element, ConstantReader annotation, BuildStep buildStep) async {
    if (Library(element.library!).last.element == element) {
      // do something for the last element
    }
  }
}
```

I would suggest if you want to use this pattern you could write it as a subclass of `Generator`

```dart
class EnumGenerator extends Generator {
  @override
  Future<String> generate(LibraryReader library, BuildStep buildStep) async {
    final lastElement = library.annotatedWith(TypeChecker.fromRuntime(Example)).last;
    // do something for the last element
  }
}
```

This PR would be a breaking change to remove the `const`. I don't think it is worth the breaking change to expose the `Iterable` for this use case.

lyledean1:
Thanks for the suggestion! Will try this out

PR #616, opened 2022-08-12
Avoid deprecated analyzer APIs

natebosch:
Bump min analyzer to `^4.4.0` which introduces some of the new APIs
used.

Switch from `ClassElement` to `InterfaceElement` in a utility method to
match the `element2` return type.

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,6 +1,6 @@
-## 1.2.3
+## 1.2.3-dev
 
-* Require `analyzer: '>=4.3.0 <5.0.0'`
+* Require `analyzer: ^4.4.0`
 * Require `sdk: '>=2.17.0 <3.0.0'`
 
 ## 1.2.2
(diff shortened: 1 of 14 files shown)


scheglov:
Note, that eventually `EnumElement` and `MixinElement` will stop implementing `ClassElement`. But I don't have enough attention to do it all in one step. So, be aware that more changes are coming next week :-)

source_gen/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
-## 1.2.3
+## 1.2.3-dev
 
-* Require `analyzer: '>=4.3.0 <5.0.0'`
+* Require `analyzer: ^4.4.0`
```
scheglov, line 3: I'm surprised that you can get away with `4.4.0`, not `4.6.0`, where we added `DartType.element2`, so not just to `InterfaceType`.

scheglov (approved):

source_gen/CHANGELOG.md
natebosch: Ah, I probably need to bump it further then.

PR #617, opened 2022-08-29
Prepare for separation of ClassElement, EnumElement and MixinElement.

scheglov:
All three implement `InterfaceElement`, but in the breaking change for `analyzer` we will stop implementing `ClassElement` in `EnumElement` and `MixinElement`. This is a minimal change necessary to make google3 green, the might be need for more changes that are not exercised there.


srawlins (approved):

jakemac53 (approved):

scheglov:
Throwing an exception explicitly seems to help.

PR #622, opened 2022-09-20
Fix another type issue, prepare for v1.2.5

kevmoo:
`root.fields` has a runtime type of `Iterable<FieldElementImpl>` which
is not compatible with the value used in the `addAll` call.

Using collection literal syntax to fix the issue

Runtime error:
`type 'ExpandIterable<InterfaceType, FieldElement>' is not a subtype of type 'Iterable<FieldElementImpl>' of 'other'`

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.2.5
+
+* Fix another issue with overly specific types.
+
 ## 1.2.4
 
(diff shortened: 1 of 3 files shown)


kevmoo:
When I sent the PR for the previous...PR...I thought "maybe I should double check there aren't other issues...nah"

So...there were other issues... 🤦

natebosch (approved):
Don't forget to land internally before publishing.

Issue #624, opened 2022-09-24
Coverage ignore option


jtdLab:
Hi,

is it possible to add an option to combining_builder  which leads to the line `//coverage:ignore-file` beeing added on top of generated files. Or is there already an existing solution?

Something like:

```yaml
targets:
  $default:
    builders:
      source_gen:combining_builder:
        options:
          coverage-ignore-file: true
```

jakemac53:
It could be added (would work similar to [ignore_for_file](https://github.com/dart-lang/source_gen#configuring-combining_builder-ignore_for_file)).

Note though that this would take effect for all source_gen builder outputs that are combined into that file, which might be fine, but is worth keeping in mind.

jtdLab:
Yep i agree a general approach would be the preferable solution.

jakemac53:
I spoke a bit with @kevmoo and it sounds like we agree this is a sensible feature to have - but I don't think either of us have the time to implement it at this time. We would accept a PR though, either with a specific option to ignore coverage (maybe `ignore_coverage: true`?) or a general `preamble` option which accepts any text and inserts it at the top of the file. If we did the more general approach we would want to ensure the text ends with a newline.

kevmoo:
See https://github.com/dart-lang/source_gen/pull/636#issuecomment-1384683749

kevmoo:
er, preamble

PR #625, opened 2022-09-28
Add preamble to CombiningBuilder

jtdLab:
Adds a `preamble` to `CombiningBuilder` as discussed in (#624).

TODO:
Wheter remove or deprecate `ignoreForFile`
Doc update

### DIFF
--- a/source_gen/lib/builder.dart
+++ b/source_gen/lib/builder.dart
@@ -32,12 +32,16 @@ Builder combiningBuilder([BuilderOptions options = BuilderOptions.empty]) {
   final ignoreForFile = Set<String>.from(
     optionsMap.remove('ignore_for_file') as List? ?? <String>[],
   );
+  final preamble = List<String>.from(
+    optionsMap.remove('preamble') as List? ?? <String>[],
+  );
   final buildExtensions =
       validatedBuildExtensionsFrom(optionsMap, _defaultExtensions);
 
   final builder = CombiningBuilder(
     includePartName: includePartName,
     ignoreForFile: ignoreForFile,
+    preamble: preamble,
     buildExtensions: buildExtensions,
   );
 
@@ -58,6 +62,8 @@ class CombiningBuilder implements Builder {
 
   final Set<String> _ignoreForFile;
 
+  final List<String> _preamble;
+
   @override
   final Map<String, List<String>> buildExtensions;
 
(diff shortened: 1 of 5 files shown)


source_gen/lib/builder.dart
```diff
@@ -32,12 +32,16 @@ Builder combiningBuilder([BuilderOptions options = BuilderOptions.empty]) {
   final ignoreForFile = Set<String>.from(
     optionsMap.remove('ignore_for_file') as List? ?? <String>[],
   );
+  final preamble = List<String>.from(
```
jakemac53, line 35: Should this just be a single string instead of a list of lines?

jtdLab: You are right it should be a single string

source_gen/test/builder_test.dart
```diff
@@ -670,6 +670,41 @@ part of a;
...
+      );
+    });
+
+    test('includes preamble if enabled', () async {
+      await testBuilder(
+        const CombiningBuilder(
+          preamble: '''
+foo
+
+bar''',
```
natebosch, line 687: [nit] Adding a comment in the preamble would be more realistic.
```suggestion
// foo

// bar''',
```

natebosch (commented):
This feature needs a CHANGELOG, and we should probably mention it in the README.

source_gen/lib/builder.dart
```diff
@@ -131,9 +137,12 @@ class CombiningBuilder implements Builder {
     final ignoreForFile = _ignoreForFile.isEmpty
         ? ''
         : '\n// ignore_for_file: ${_ignoreForFile.join(', ')}\n';
+
+    final preamble = _preamble.isEmpty ? '' : '\n$_preamble\n';
+
     final output = '''
 $defaultFileHeader
-${languageOverrideForLibrary(inputLibrary)}$ignoreForFile
+${languageOverrideForLibrary(inputLibrary)}$ignoreForFile$preamble
```
jakemac53, line 145: There is some question as to exactly where this should go, but I think probably after the ignore for file lines is fine?

We should mention in the docs that this is where it will go though

jtdLab: You could argue for `preamble` as the more general option beeing first and `ignoreForFile` as the more specific beeing second. But its a matter of taste i guess

jakemac53: yeah I don't think it matters much either way? we should just document which is which :)

jtdLab: Done

jakemac53 (approved):

jakemac53:
thanks @jtdLab ! This is published now

PR #626, opened 2022-10-25
update the pubspec 'repository' field

devoncarew:
- update the pubspec 'repository' field

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.2.7-dev
+
+* Update the value of the pubspec `repository` field.
+
 ## 1.2.6
 
 * Add support for a `preamble` option to `combining_builder`.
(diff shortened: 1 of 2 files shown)


natebosch (approved):
If this is worth publishing by itself, it's OK for it to go out without rolling the package internally since these files have no impact in bazel. You could skip the `-dev` if you plan to publish.

devoncarew:
> If this is worth publishing by itself, it's OK for it to go out without rolling the package internally since these files have no impact in bazel. You could skip the `-dev` if you plan to publish.

Will do!

It looks like the CI is unhappy w/ some new analyzer deprecations from `5.2.0`. I'll hold off on landing this until I've had a chance to look into the deprecations.

devoncarew:
It looks like in order to make the CI happy, we either need to address the deprecations (just referencing the non-deprecated members) and set analyzer `5.2.0` - the current version - as our min. required version, or adjust the analysis options to ignore deprecations.

natebosch:
https://github.com/dart-lang/source_gen/pull/627 bumps the analyzer dep and migrates from deprecated APIs.

devoncarew:
I'll update this once the other PR lands.

PR #647, opened 2023-02-14
Add support for `build_extensions` configuration of builders producing multiple files

d-markey:
The `validatedBuildExtensionsFrom()` in `utils.js` currently rejects `build_extensions` configurations that use a list of strings rather than a single string.

This PR adds support for lists of strings.

### DIFF
--- a/source_gen/lib/src/utils.dart
+++ b/source_gen/lib/src/utils.dart
@@ -206,7 +206,7 @@ Map<String, List<String>> validatedBuildExtensionsFrom(
       if (o is! String || (i == 0 && !o.endsWith('.dart'))) {
         throw ArgumentError(
           'Invalid output extension `${entry.value}`. It should be a string '
-          'or a list of strings ending with `.dart`',
+          'or a list of strings with the first ending with `.dart`',
         );
       }
     }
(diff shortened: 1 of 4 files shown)


source_gen/lib/src/utils.dart
```diff
@@ -195,15 +200,19 @@ Map<String, List<String>> validatedBuildExtensionsFrom(
...
-        'string ending with `.dart`',
-      );
+    final output = (entry.value is List) ? entry.value as List : [entry.value];
+
+    for (var i = 0; i < output.length; i++) {
+      final o = output[i];
+      if (o is! String || (i == 0 && !o.endsWith('.dart'))) {
+        throw ArgumentError(
+          'Invalid output extension `${entry.value}`. It should be a string '
+          'or a list of strings ending with `.dart`',
```
natebosch, line 210:
```suggestion
          'Invalid output extension `${entry.value}`. It should be a string '
          'or a list of strings with the first ending with `.dart`',
```

natebosch (approved):

d-markey:
Hello, thanks for the merge, I'd like to know also if you have any timeline in mind for releasing an update on pub.dev?

PR #659, opened 2023-04-25
Fix issue with `isExactlyType` that resulted in unexpected

benjross:
"Null check operator used on a null value" errors.


natebosch:
It looks like you have your fork configured to reject pushes from the upstream maintainers.

You can apply the changes from these commits to get the CI green.
https://github.com/dart-lang/source_gen/compare/master...fix-issue

natebosch (approved):

PR #664, opened 2023-05-04
Use prefixed tags for publish config

natebosch:
The auto-publish comment already indicates that the package name prefix
should be used.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -6,7 +6,7 @@ on:
   pull_request:
     branches: [ master ]
   push:
-    tags: [ 'v[0-9]+.[0-9]+.[0-9]+*' ]
+    tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+*' ]
 
 jobs:
   publish:


kevmoo (approved):

devoncarew:
We really only need to use prefix tags for packages in mono repos; `v1.2.3` for single package repos, and `my_name-v1.2.3` for packages in mono-repos (https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose#mono-repos).

natebosch:
> We really only need to use prefix tags for packages in mono repos

How does it pick the "Publish Tag' it adds to the comments - that is showing with the package prefix.

We've been using a package prefix in this repo for a while now, though I don't remember the reason why we started that pattern.

devoncarew:
It figures out whether its in a mono repo or not, and expects one form of the tag or the other:

https://github.com/dart-lang/ecosystem/blob/main/pkgs/firehose/lib/src/repo.dart#L61

We can make this configurable if necessary - we'd just have to figure out the best way to do that (a yaml file in the repo? an arg to the publish task - https://github.com/dart-lang/source_gen/blob/master/.github/workflows/publish.yaml?)

PR #666, opened 2023-05-09
Prepare for adding InvalidType, make 'TypeChecker.isAssignableFromType()' null safe.

scheglov:
Prepares for landing https://dart-review.googlesource.com/c/sdk/+/300181

I had to make this check in cl/529250286, so I guess there is a type without an element that gets passed, most probably the newly added `InvalidType`. Anyway, the code was unsafe.


natebosch (approved):

kevmoo:
@scheglov - I'm hitting it with a record!

PR #669, opened 2023-06-01
Require Dart 3, update lints

kevmoo:


source_gen/lib/src/utils.dart
```diff
@@ -64,18 +64,12 @@ String urlOfElement(Element element) => element.kind == ElementKind.DYNAMIC
...
-      return normalizeDartUrl(url);
-    case 'package':
-      return packageToAssetUrl(url);
-    case 'file':
-      return fileToAssetUrl(url);
-    default:
-      return url;
-  }
-}
+Uri normalizeUrl(Uri url) => switch (url.scheme) {
```
natebosch, line 67: 👍

natebosch (approved):

source_gen/lib/src/utils.dart
kevmoo: I can't get over how nice it is!

source_gen/lib/src/library.dart
```diff
@@ -98,10 +98,7 @@ class LibraryReader {
   /// * `asset`
   ///
   /// May throw [ArgumentError] if it is not possible to resolve a path.
-  Uri pathToUrl(dynamic toUrlOrString) {
```
natebosch, line 101: This was statically breaking.

This is worth doing, but we might have to wait for a major version bump, or add a `dart fix` fix for it.

natebosch: https://github.com/dart-lang/source_gen/pull/673

kevmoo: annoying!

PR #670, opened 2023-06-20
Add a `node` to `InvalidGenerationSourceError`

natebosch:
Allow attempting to find the source location from an AstNode as an
alternative to finding it from an Element.

### DIFF
--- a/source_gen/lib/src/span_for_element.dart
+++ b/source_gen/lib/src/span_for_element.dart
@@ -58,7 +58,7 @@ SourceSpan spanForNode(AstNode node) {
   final unit = node.thisOrAncestorOfType<CompilationUnit>()!;
   final element = unit.declaredElement!;
   final contents = element.source.contents.data;
-  final url = element.source.uri;
+  final url = assetToPackageUrl(element.source.uri);
   final file = SourceFile.fromString(contents, url: url);
   return file.span(node.offset, node.offset + node.length);
 }
--- a/source_gen/lib/src/span_for_element.dart
+++ b/source_gen/lib/src/span_for_element.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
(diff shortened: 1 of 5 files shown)


jakemac53 (approved):

source_gen/lib/src/span_for_element.dart
```diff
@@ -51,3 +52,13 @@ SourceSpan spanForElement(Element element, [SourceFile? file]) {
 
   return file.span(element.nameOffset, element.nameOffset + element.nameLength);
 }
+
+/// Returns a source span that spans the location where [node] is written.
+SourceSpan spanForNode(AstNode node) {
+  final unit = node.thisOrAncestorOfType<CompilationUnit>()!;
+  final element = unit.declaredElement!;
```
natebosch, line 59: This seems to be fragile in with the `build_test` use of `Resolver`. Depending on which specific member I read this `CompilationUnit` seems to sometimes have a null `declaredElement`.

My hunch is that in practical use cases - those where the code generator is working with the AST model instead of the element model - this will be more reliable than the `build_test` stuff, but I only know for sure that it works in the example that surfaced a request for this.

I also worry if this has a bug for nodes from part files.

@scheglov - Do you know if there is a more reliable way to get from an `AstNode` to the full content of the source file that the node was read from?

scheglov: Most `AstNode`s come from a `ResolvedUnitResult`, which has `String get content`.

I never liked `Source.contents`, but we still have it by historical reasons. What it does, it read the current content of the file, not necessary the same that was used to produce the node, because of race conditions.

And then there are AST nodes that are constants exposed from element model, these are disconnected from a `CompilationUnit`, so the null assert will fail.

natebosch: I explored this a bit and I think it could get a little unwieldy to try to do this the "right way".

In the known use case for this code path we do have a `ResolvedUnitResult`, so we _could_ take a `(ResolvedUnitResult, AstNode)` argument, or as two separate arguments. I think it

PR #681, opened 2023-09-25
Use type argument for Future of nullable value in test

srawlins:
This is a weird one, so I'm going to try to carefully explain the situation and my steps. Here is the code in question:

```dart
void _testSimpleValue(String testName, Object? value, Object? expected) {
  // ...
  _testFunction('Future<$testName>', Future.value(value), expected);
  // -------------------------------------^ inferred type argument
  // ...
}

void _testFunction(String testName, Object? value, Object? expected) { /* ... */ }
```

This code works today. It appears that the type of `Future.value(...)` inferred by CFE is `Future<Object?>`. A `Future<Object?>` constructed with the `value` constructor, and an `Object?` argument is sound. However, in https://github.com/dart-lang/language/issues/3304, it was decided that CFE infers the wrong type here, and should infer `Future<Object>`.

If and when CFE will change its behavior remains to be seen, but analyzer [will start reporting](https://github.com/dart-lang/sdk/issues/53253) an issue that an `Object?` argument was passed to a `Future.value` constructor for an argument type of `Object`, and that would be invalid. The situation there is that `Future.value`'s single parameter is optional, so it's type is `T?`, but at runtime a null value will throw an exception, so we can pretend the parameter is both required and non-nullable when `T` is non-nullable.

So to prepare for a CFE change, and support analyzer's new warning, we can just put an explicit type argument.

---

<details>

</details>

### DIFF
--- a/source_gen/test/output_helpers_test.dart
+++ b/source_gen/test/output_helpers_test.dart
@@ -66,7 +66,7 @@ void _testSimpleValue(String testName, Object? value, Object? expected) {
 
   assert(value is! Future);
 
-  _testFunction('Future<$testName>', Future.value(value), expected);
+  _testFunction('Future<$testName>', Future<Object?>.value(value), expected);
 
   if (value is Iterable) {
     _testFunction(


jakemac53 (approved):
Unfortunate, but `Future.value` is just a horrible API unfortunately. We should have fixed it in Dart 3.0 😞 .

PR #687, opened 2023-10-03
Make InvalidGenerationSourceError an Exception

natebosch:
See https://github.com/dart-lang/build/issues/3585

Rename to `InvalidGenerationSource` and change to a subtype of
`Exception`. This class is not thrown to indicate a programming error in
the code that is running, it is thrown to indicate an error in the code
under analysis. It was originally implemented without a super type and
the name ending in "Error" was not specifically discussed, but it was
intended to convey an "error in build input", not an error in the
builder implementation (where it is thrown). Later a lint required it to
be a subtype of either `Error` or `Exception`, and the `Error` supertype
was chosen without discussion because of the name, even though it's not
thrown as an `Error` in the sense where that distinction matters.

Add a type alias for the exception with the old name. This can be
deprecated whenever we have the bandwidth to handle the cleanup, but it
will not be deprecated immediately.

Change from `extends Error` to `implements Exception`. This is
technically breaking, but it is unlikely to make a significant impact in
real world usage scenarios. It may impact how the error surfaces to end
users in some places.

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -1,4 +1,8 @@
-## 1.4.1-wip
+## 1.5.0-wip
+
+- Rename `InvalidGenerationSourceError` to `InvalidGenerationSource`. Change
+  from a subtype of `Error` to a subtype of `Exception`. This may be breaking if
+  a builder relies on a `on Exception catch` to ignore this error.
 
 ## 1.4.0
 
--- a/source_gen/lib/source_gen.dart
+++ b/source_gen/lib/source_gen.dart
@@ -6,7 +6,8 @@ export 'src/builder.dart'
     show LibraryBuilder, PartBuilder, SharedPartBuilder, defaultFileHeader;
 export 'src/constants/reader.dart' show ConstantReader;
 export 'src/constants/revive.dart' show Revivable;
-export 'src/generator.dart' show Generator, InvalidGenerationSourceError;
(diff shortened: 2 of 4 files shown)


natebosch:
@kevmoo - do you have concerns about the potential breakage cause by changing an `extends Error` to `implements Exception`?

natebosch:
This change did have a successful global presubmit.

kevmoo:
> @kevmoo - do you have concerns about the potential breakage cause by changing an `extends Error` to `implements Exception`?

Not really? What are the failure modes here?

natebosch:
> What are the failure modes here?

The only thing I can think of is something like a builder may have an outer `try/catch on Exception` or zone error handler which they expect to not catch `InvalidGenerationSource` (because that already has a user friendly toString that they want to expose).

The context is that we are going to start showing stack traces (even without `-v`) for caught `Error`, because we expect most of those to be bugs in the builder implementation, and the stack trace can be helpful in bug reports.

Some builders use `InvalidGenerationSourceError` (correctly) to report issues with the input code, and we don't want those to show a stack trace by default.

natebosch:
> an outer `try/catch on Exception` or zone error handler which they expect to not catch `InvalidGenerationSource`

Even in these cases, the worst I expect is potentially changed error reporting back to the user. It's obviously not likely to change any happy path behavior.

PR #703, opened 2024-03-10
Support path with parenthesis

hypen-flutter:
Parenthesized path like `(this)`
---
I am creating a nextjs-like file-based routing system. It contains a path grouping mechanism that uses the `(parenthesized)` path. `source_gen` currently depends on `glob`, and the `glob` uses the regular expression. In regular expression, `()` works as a capture. Thus, if the path contains the `(this)`-like path, `source_gen` throws an error. 

image

To support this path, I change the code that appends the escaping character `\` before the parenthesis. 

This is a one-line change, so I did not add the test, modify the changelog, or change the pubspec package version. If those are required, please let me know. 

<details>

</details>

### DIFF
--- a/source_gen/lib/builder.dart
+++ b/source_gen/lib/builder.dart
@@ -85,8 +85,7 @@ class CombiningBuilder implements Builder {
     final pattern = buildStep.inputId
         .changeExtension('.*$_partFiles')
         .path
-        .replaceAll('(', r'\(')
-        .replaceAll(')', r'\)');
+        .replaceAllMapped(RegExp(r'\((.*)\)'), (match) => '\\(${match[1]}\\)');
 
     final inputBaseName =
         p.basenameWithoutExtension(buildStep.inputId.pathSegments.last);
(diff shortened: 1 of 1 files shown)


source_gen/lib/builder.dart
```diff
@@ -82,7 +82,10 @@ class CombiningBuilder implements Builder {
   @override
   Future<void> build(BuildStep buildStep) async {
     // Pattern used for `findAssets`, which must be glob-compatible
-    final pattern = buildStep.inputId.changeExtension('.*$_partFiles').path;
+    final pattern = buildStep.inputId
```
jakemac53, line 85: I think we should probably do `Glob.quote` here instead? That will handle other characters as well.

jakemac53:
> This is a one-line change, so I did not add the test, modify the changelog, or change the pubspec package version. If those are required, please let me know.

We should add a test, but the version can stay as is. An entry in the changelog should be added for the `1.5.1-wip` version which will be the next release.

hypen-flutter:
I've added a test and modified the changelog.

source_gen/lib/builder.dart
hypen-flutter: Thanks for your help! I didn't know it existed.

jakemac53 (approved):

PR #711, opened 2024-05-29
update lints dependency

kevmoo:


source_gen/pubspec.yaml
```diff
@@ -21,6 +21,6 @@ dev_dependencies:
   _test_annotations:
     path: ../_test_annotations
   build_test: ^2.0.0
-  dart_flutter_team_lints: ^2.0.0
+  dart_flutter_team_lints: ^3.1.0
```
devoncarew, line 24: FYI, if we use a constraint of `^3.0.0` here, we only need to bump the min. sdk constraint for this package to `3.1.0`.

devoncarew (approved):

source_gen/pubspec.yaml
kevmoo: Yeah, but I'm a big believer in moving quickly to the future. There aren't a lot of big fixes here that will restrict folks.

PR #713, opened 2024-06-04
Revive to the object's `variableElement` if available

mattrberry:
Variables have been tracked on DartObject since https://dart.googlesource.com/sdk/+/54b7f4b72a1701f8f9a0334c94ce6f59732bd261. This change uses the variable in the reviver if it exists. I believe the existing tests in https://github.com/dart-lang/source_gen/blob/master/source_gen/test/constants_test.dart already cover this, but lmk if you'd like any additional tests

- Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

---

<details>

</details>

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -4,6 +4,7 @@
   `GeneratorForAnnotation.generateForAnnotatedElement`.
 - Support all the glob quotes.
 - Require Dart 3.4.0
+- Revive to the object's `variableElement` if available
 
 ## 1.5.0
 
--- a/source_gen/lib/src/constants/revive.dart
+++ b/source_gen/lib/src/constants/revive.dart
@@ -20,6 +20,21 @@ import '../utils.dart';
 /// Dart source code (such as referencing private constructors). It is up to the
 /// build tool(s) using this library to surface error messages to the user.
 Revivable reviveInstance(DartObject object, [LibraryElement? origin]) {
+  final variableElement = object.variable;
+  if (variableElement != null &&
+      variableElement.isConst &&
+      variableElement.isPublic) {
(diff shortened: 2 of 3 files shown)


kevmoo:
I mean, if this changes the behavior, we should have a test that shows the behavior change/improvement...right?

mattrberry:
This is really just a fast-path that avoids iterating all members in the library rather than a change in behavior.. That said, `revive` doesn't support extension types yet, so this technically does change behavior in that it adds support for extension types assigned to variables that would be otherwise un-revivable. It's not really a regression test for this feature, but it's something.

kevmoo:
Please add a changelog entry!

kevmoo (approved):

jakemac53:
Fwiw the issue seems to be that imports don't get added for these variables in generated mockito code. That might just be a mockito bug or something I am not sure.

mattrberry:
I also just looked at one case and it seems like the source_gen Revivable is correct, which would align with your thought that mockito just isn't emitting the import. I haven't worked much with mockito, but I can take a look if you aren't already

jakemac53:
> but I can take a look if you aren't already

If you can that would be great, I only looked enough to understand mockito was the problem

mattrberry:
Mockito [assumes](https://github.com/dart-lang/mockito/blob/2259e105b335b7a4b3a4df313ce4099e8a92a959/lib/src/builder.dart#L1874) that the import uri is where the object's type is defined, which isn't always true. This seems like a bug in mockito.

e.g.

```dart
# foo.dart
class Foo {
  static final bar = Bar();
}
# bar.dart
class Bar {}
```

With this change, we now create a `Revivable` for `Foo.bar`, and mockito reads the import for that as `bar.dart` rather than `foo.dart`.

PR #717, opened 2024-09-10
Require Dart 3.5, move to workspaces

kevmoo:


jakemac53:
Hah! I was just touching this repo to migrate to the new formatter and was going to do this, but thanks for beating me to it 👍

jakemac53 (approved):

jakemac53:
We should also add the package health checks (but separately)

PR #722, opened 2024-09-23
Require analyzer ^6.9.0, switch to using 'enclosingElement3'.

scheglov:
Prepare for https://dart-review.googlesource.com/c/sdk/+/385100

### DIFF
--- a/source_gen/pubspec.yaml
+++ b/source_gen/pubspec.yaml
@@ -6,7 +6,7 @@ repository: https://github.com/dart-lang/source_gen/tree/master/source_gen
 resolution: workspace
 
 environment:
-  sdk: ^3.6.0-217.0.dev
+  sdk: ^3.6.0-270.0.dev
 
 dependencies:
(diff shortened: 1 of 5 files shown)


source_gen/pubspec.yaml
```diff
@@ -6,10 +6,10 @@ repository: https://github.com/dart-lang/source_gen/tree/master/source_gen
 resolution: workspace
 
 environment:
-  sdk: ^3.5.0
+  sdk: ^3.6.0-edge
```
jakemac53, line 9: Instead of `-edge` we ideally want a dev release, specifically `3.6.0-217.0.dev` it looks like.

scheglov:
Is there something else to fix, or should I land it?

kevmoo:
You'll have to bump min SDK on `example_usage` as well and re-run monorepo generate

jakemac53 (approved):

PR #725, opened 2024-10-09
Add dart format width comments

### LINKED ISSUE #724: Add "// dart format width=80" comment in generated code
After many years of users asking for it, we are adding support to the formatter for project-wide page width configuration using a surrounding `analysis_options.yaml` file (https://github.com/dart-lang/sdk/issues/56863). However, tools like source_gen don't necessarily have access to (or want to access) IO or know where generated code will end up on disk.

To avoid breaking CI by having the code get generated and formatted at one width and then get formatted to a different width when the user runs `dart format`, we also added support for putting a marker comment in a file to specify an explicit width, like:

```dart
// dart format width=80
```

When this comment is present at the top of the file (before any code, but can be after other comments), then that comment will take precedence and the file will reliably be formatted at that width regardless of any surrounding configuration. So if code generators insert that comment at the top of the code they generate, they don't have to worry about reading the user's configuration.

This issue is requesting that source_gen be updated to add that comment to the code it generates. Since it's just a comment, it's safe to add this at any time. Let me know if you have any questions/feedback and I'll do my best.

natebosch:
Closes #724

The 3 places where a Dart unit can be output are as a `LibraryBuildr`,
as a `PartBuilder`, or as a combined part. `SharedPartBuilder` output
does not need the width comment. Add a constant for the comment and use
it while writing each of the 3 types of output.

### DIFF
--- a/example_usage/lib/library_source.g.dart
+++ b/example_usage/lib/library_source.g.dart
@@ -1,4 +1,5 @@
 // GENERATED CODE - DO NOT MODIFY BY HAND
+// dart format width=80
 
 // ignore_for_file: lint_a, lint_b
 
--- a/example_usage/lib/library_source.info.dart
+++ b/example_usage/lib/library_source.info.dart
@@ -1,4 +1,5 @@
 // GENERATED CODE - DO NOT MODIFY BY HAND
+// dart format width=80
 
 // **************************************************************************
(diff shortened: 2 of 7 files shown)


munificent (approved):
Thank you for the quick turnaround!

PR #727, opened 2024-10-09
Don't use workspace resolution for example

natebosch:
Use non-workspace resolution because `build_runner` does not support
packages where the `pubspec.lock` file is not at the package root.

See https://github.com/dart-lang/build/issues/3721

### DIFF
--- a/example_usage/pubspec.yaml
+++ b/example_usage/pubspec.yaml
@@ -1,6 +1,8 @@
 name: source_gen_example_usage
 publish_to: none
-resolution: workspace
+# TODO(https://github.com/dart-lang/build/issues/3721) - build_runner broken by
+# workspace resolution.
+# resolution: workspace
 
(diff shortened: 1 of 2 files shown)


natebosch:
Why does this work on CI when it validated the build https://github.com/dart-lang/source_gen/actions/runs/11246229813/job/31267721145?pr=725

But it doesn't work for me locally....

jakemac53:
@natebosch do you have any `.packages` files hanging around? This can cause things to get super weird and I wonder if its related.

natebosch:
> do you have any `.packages` files hanging around?

Yes, this was exactly the issue.

jakemac53:
Fwiw I filed https://github.com/dart-lang/package_config/issues/158 which is the root cause here. cc @lrhn we should discuss making this change

PR #728, opened 2024-10-10
Read parts with the defining compilation unit

natebosch:
The direct read of parts from the library element is deprecated.
Retain ignores for reading imports and exports since they
will be restored without a deprecation.

Unblocks https://dart-review.googlesource.com/c/sdk/+/388643


jakemac53:
It actually sounds like these are getting undeprecated... https://dart-review.googlesource.com/c/sdk/+/389142

scheglov (approved):

PR #729, opened 2024-10-11
Read libraryImports/libraryExports from the defining unit.

scheglov:
Follow up for https://github.com/dart-lang/source_gen/pull/728

### DIFF
--- a/source_gen/lib/src/library.dart
+++ b/source_gen/lib/src/library.dart
@@ -37,10 +37,8 @@ class LibraryReader {
   Iterable<Element> get allElements => [
         element,
         ...element.topLevelElements,
-        // ignore: deprecated_member_use
-        ...element.libraryImports,
-        // ignore: deprecated_member_use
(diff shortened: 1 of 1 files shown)


source_gen/lib/src/library.dart
```diff
@@ -37,10 +37,8 @@ class LibraryReader {
   Iterable<Element> get allElements => [
         element,
         ...element.topLevelElements,
-        // ignore: deprecated_member_use
-        ...element.libraryImports,
-        // ignore: deprecated_member_use
-        ...element.libraryExports,
+        ...element.definingCompilationUnit.libraryImports,
```
kevmoo, line 40: are the SDK and pkg:analyzer constraints all valid?

scheglov: `analyzer 6.9.0` has `element.definingCompilationUnit.libraryImports`

kevmoo (approved):
thanks!

PR #733, opened 2024-12-10
Only add dart format width comment on format

### LINKED ISSUE #724: Add "// dart format width=80" comment in generated code

natebosch:
Closes #724

A use case with a custom `formatCode` callback can reasonably use a
different line width than the default, so only add the width comment
when this package is responsible for performing a format.

Uses with a custom header will no longer slot the width comment in
following the header.

Some outputs will no longer have the comment. This is not breaking since
there has been no published since the comment was introduced.


jakemac53 (approved):

kevmoo:
So the idea here is that we don't have a good way to extract the style info from analysis_options, so we're just going to hard-wire?

natebosch:
> So the idea here is that we don't have a good way to extract the style info from analysis_options, so we're just going to hard-wire?

Yes. It is not guaranteed that the configuration is in a file that we make readable during builds.

marvin-kolja:
@natebosch, may I ask why the comment was added as a first line? This conflicts with GitHub Linguist to recognize the file as generated (https://github.com/github-linguist/linguist/blob/main/lib/linguist/generated.rb#L737). I was about to fix a related issue in Freezed (https://github.com/rrousselGit/freezed/issues/1261) when I came across this.

natebosch:
I'm not sure what it would take to change the ecosystem to allow the comment on the second line. If we need to reorder the lines we either need a breaking change to the signature of the callback (we'd need to pass the header and body separately to put the line width comment between them) or we can do something hacky and hardcode support for the generate content line specifically. I'd lean towards doing the latter so I'll open a PR with that approach, but I haven't fully thought it through and I can't roll out the change. @kevmoo @jakemac53 if you think this is urgent you may want to run with https://github.com/dart-lang/source_gen/pull/760

marvin-kolja:
Thanks for the quick response.

_For now, I've also opened a request to improve how GitHub Linguist recognizes generated files by looking at the first 3 lines https://github.com/github-linguist/linguist/pull/7448._

Like that second idea, I will have a look.

PR #735, opened 2024-12-16
Add APIs to expose analyzer elements2 models.

scheglov:
See https://docs.google.com/presentation/d/1Vxr-y5ljuo0Os78R4tKZg8cc0In984EQcLs06HOqS28/edit#slide=id.p for the new element model details. It is possible that you will not able access it, unless you work for Google.

### DIFF
--- a/source_gen/lib/source_gen.dart
+++ b/source_gen/lib/source_gen.dart
@@ -10,6 +10,6 @@ export 'src/generator.dart'
     show Generator, InvalidGenerationSource, InvalidGenerationSourceError;
 export 'src/generator_for_annotation.dart' show GeneratorForAnnotation;
 export 'src/library.dart' show AnnotatedElement, LibraryReader;
-export 'src/span_for_element.dart' show spanForElement;
+export 'src/span_for_element.dart' show spanForElement, spanForElement2;
 export 'src/type_checker.dart' show TypeChecker, UnresolvedAnnotationException;
 export 'src/utils.dart' show typeNameOf;
--- a/source_gen/lib/src/generator_for_annotation.dart
+++ b/source_gen/lib/src/generator_for_annotation.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/element2.dart';
 import 'package:build/build.dart';
 
 import 'constants/reader.dart';
@@ -58,8 +59,13 @@ abstract class GeneratorForAnnotation<T> extends Generator {
       typeChecker,
       throwOnUnresolved: throwOnUnresolved,
     )) {
-      final generatedValue = generateForAnnotatedElement(
+      var generatedValue = generateForAnnotatedElement(
         annotatedElement.element,
+        annotatedElement.annotation,
+        buildStep,
+      );
+      generatedValue ??= generateForAnnotatedElement2(
+        annotatedElement.element2,
         annotatedElement.annotation,
         buildStep,
       );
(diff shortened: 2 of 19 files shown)


jakemac53 (approved):

source_gen/test/generator_for_annotation_test.dart
```diff
@@ -142,47 +142,6 @@ $dartFormatWidth
     );
   });
 
-  test('applies to annotated directives', () async {
```
kevmoo, line 145: What's the story here?

source_gen/pubspec.yaml
```diff
@@ -26,3 +27,31 @@ dev_dependencies:
   dart_flutter_team_lints: ^3.1.0
   term_glyph: ^1.2.0
   test: ^1.16.0
+
```
kevmoo, line 30: just published these! we can drop 'em

source_gen/lib/source_gen.dart
```diff
@@ -10,6 +10,6 @@ export 'src/generator.dart'
     show Generator, InvalidGenerationSource, InvalidGenerationSourceError;
 export 'src/generator_for_annotation.dart' show GeneratorForAnnotation;
 export 'src/library.dart' show AnnotatedElement, LibraryReader;
-export 'src/span_for_element.dart' show spanForElement;
+export 'src/span_for_element.dart' show spanForElement, spanForElement2;
```
kevmoo, line 13: Need a changelog update

source_gen/test/generator_for_annotation_test.dart
scheglov: Directives in the element model are no longer elements, so we cannot support annotations on them. Or at least the way it used to work. This so far did not cause any breakage in google3, but I have not yet migrated external packages, only a lot of internal generators.

source_gen/lib/src/generator_for_annotation.dart
```diff
@@ -93,5 +99,28 @@ abstract class GeneratorForAnnotation<T> extends Generator {
...
+  /// returned through an iterable or stream they will be deduplicated.
+  /// Typically each value will be an independent unit of code and the
+  /// deduplication prevents re-defining the same member multiple times. For
+  /// example if multiple annotated elements may need a specific utility method
+  /// available it can be output for each one, and the single deduplicated
+  /// definition can be shared.
+  ///
+  /// Implementations should return `null` when no content is generated. Empty
+  /// or whitespace-only [String] instances are also ignored.
+  dynamic generateForAnnotatedElement2(
```
kevmoo, line 121: Wait, so which version is going to be called?

source_gen/pubspec.yaml
jakemac53: These are actually needed, they expose a different API to get at the new element models, but its only to help with the internal migration.

source_gen/lib/source_gen.dart
scheglov: It will never be published, we will restore the name with different types.

source_gen/lib/src/generator_for_annotation.dart
scheglov: Both, but any specific generator will override only one.
So, both migrated and not-yet-migrated generators will work.

source_gen/test/generator_for_annotation_test.dart
scheglov: I asked in our [chat](https://chat.google.com/room/AAAA-IuSyjA/0e1s6XkTmCA/0e1s6XkTmCA?cls=10).

PR #754, opened 2025-04-22
Ignore analyzer deprecations (for now)

kevmoo:


analysis_options.yaml
```diff
@@ -5,6 +5,9 @@ analyzer:
     strict-casts: true
     strict-inference: true
     strict-raw-types: true
+  errors:
+    # analyzer deprecations
```
natebosch, line 9: What version are they deprecated in? I don't see any warnings with the latest published, do I need a git override to repro?

natebosch: Nevermind, I'm just not getting diagnostics from the anlaysis server in this repo.

kevmoo: there are very much deprecation warnings w/ the latest stable release!

natebosch (approved):

PR #765, opened 2025-07-08
skip files with only known annotation names

jakemac53:
Previously we would bail out early for files with no top level annotations, this also skips well known annotations (although it isn't perfect, and assumes people aren't shadowing these names).

### DIFF
--- a/source_gen/CHANGELOG.md
+++ b/source_gen/CHANGELOG.md
@@ -6,6 +6,8 @@
   For questions please use https://github.com/dart-lang/build/discussions.
 - Updated the minimum package versions for a number of dependencies.
 - Require Dart 3.7.0
+- Update the GeneratorForAnnotation optimization to skip files with well known
+  annotation names such as `override`, `Deprecated`, and `pragma`.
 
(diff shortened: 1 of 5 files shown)


source_gen/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   For questions please use https://github.com/dart-lang/build/discussions.
 - Updated the minimum package versions for a number of dependencies.
 - Require Dart 3.7.0
+- Update the GeneratorForAnnotation optimization to skip files with well known
```
davidmorgan, line 9: -dev is released, I guess start a block for `3.0.0`

davidmorgan (approved):
Thanks!

Some tests are failing--I guess because they are checking failures and so have no annotation, so the fix should be to add an annotation :) PTAL.

source_gen/CHANGELOG.md
jakemac53: Updated it to `-wip` following the standard in other repos for unreleased in progress versions.

Issue #775, opened 2025-08-05
Stop using `dart:mirrors` to allow AOT compilation


davidmorgan:
For performance we'd like to compile builders with AOT, but this doesn't work with mirrors.

I suggest to deprecated `TypeChecker.fromRuntime` and add the closest replacement we can do without mirrors: `TypeChecker.typeNamed(Type type, {String? inPackage})`. It takes the type name from `Type` and optionally a package name. If you want full URL then use `fromUrl` instead.

Release with deprecated+suggested replacements then remove with a major version bump.

Note that AOT support in build itself needs some work, but I plan on tackling that soon :) so it'd be good to get the source_gen change underway.

PR #776, opened 2025-08-05
Deprecate TypeChecker.fromRuntime, add TypeChecker.typeNamed as recommended replacement.

davidmorgan:
For #775

### DIFF
--- a/source_gen/lib/src/generator_for_annotation.dart
+++ b/source_gen/lib/src/generator_for_annotation.dart
@@ -42,7 +42,15 @@ import 'type_checker.dart';
 /// may be helpful to check which elements have a given annotation.
 abstract class GeneratorForAnnotation<T> extends Generator {
   final bool throwOnUnresolved;
+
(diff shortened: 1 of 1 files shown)


source_gen/lib/src/generator_for_annotation.dart
```diff
@@ -42,12 +42,25 @@ import 'type_checker.dart';
 /// may be helpful to check which elements have a given annotation.
 abstract class GeneratorForAnnotation<T> extends Generator {
   final bool throwOnUnresolved;
+  final String? inPackage;
```
kevmoo, line 45: Doc comments here, please? So it's clear how they are used.

kevmoo (changes requested):

source_gen/lib/src/generator_for_annotation.dart
davidmorgan: Done.

PR #778, opened 2025-08-07
Release 3.1.0.

davidmorgan:
Analyzer 8 won't version solve yet due to build+dart_style, I tested it with a dep override here

https://github.com/dart-lang/source_gen/pull/777

and added ignores + one tiny fix, copied here.

So this should be good to go on 7+8 now :)


kevmoo (changes requested):
Please update the example constraint, too!

kevmoo (approved):
I gottcha!

PR #782, opened 2025-08-17
Bring up to analyzer 8.1.1, get rid of deprecated classes and methods.

fshcheglov:
Also deleted
```dart
    if (element case final Annotatable annotatable) {
      final annotations = annotatable.metadata2.annotations;
      if (annotations.isEmpty) {
        return null;
      }
    }
  ```
  from `source_gen/lib/src/type_checker.dart` because it repeated a check that was done a few lines later.

### DIFF
--- a/example/lib/src/multiplier_generator.dart
+++ b/example/lib/src/multiplier_generator.dart
@@ -2,9 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// ignore_for_file: deprecated_member_use until analyzer 7 support is dropped.
-
-import 'package:analyzer/dart/element/element2.dart';
+import 'package:analyzer/dart/element/element.dart';
 import 'package:build/build.dart';
 import 'package:source_gen/source_gen.dart';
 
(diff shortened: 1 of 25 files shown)


scheglov:
I'd like to remove deprecated APIs from the analyzer, and for this there should be no clients of these deprecated APIs in google3. For clients that are developed outside of google3, the ideal path is to land changes externally, and then roll into google3. But if you are hesitant to make external changes, and OK with applying the same changes directly in google3, this would work for me too.

davidmorgan:
> I'd like to remove deprecated APIs from the analyzer, and for this there should be no clients of these deprecated APIs in google3. For clients that are developed outside of google3, the ideal path is to land changes externally, and then roll into google3. But if you are hesitant to make external changes, and OK with applying the same changes directly in google3, this would work for me too.

source_gen is not currently synced into google3, so landing here won't help you directly. It's fine for me to land the changes directly in google3.


> If functionality is not being changed, I'm fine bumping a minor version to require the latest pkg:analyzer.

The ecosystem is not yet fully on analyzer 8, we need to fix up and release all our generators and let others do the same. In parallel to that we want users of source_gen to remove their use of mirrors. So I'm suggesting we should do the mirrors-removing release before the analyzer-7-removing release in order to give them an easier/smoother time.

Let me actually send the PR removing use of mirrors, then we can release and get _that_ started and then land this one :)

davidmorgan:
Okay, sent remove mirrors

https://github.com/dart-lang/source_gen/pull/783

let's land+release that then land+release this as 4.1.0.

google3 will need more work because it needs to migrate onto the API without mirrors, I will aim take a look at that in the next week.

kevmoo:
We good here @davidmorgan ?

davidmorgan (approved):

PR #785, opened 2025-08-27
Tweak publish workflow.

davidmorgan:
The publish workflow doesn't seem to be working, this makes it more like the `build` repo publish workflow which might help :)

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -5,6 +5,7 @@ name: Publish
 on:
   pull_request:
     branches: [ master ]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
   push:
     tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+*' ]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -5,13 +5,16 @@ name: Publish
...
+    types: [opened, synchronize, reopened, labeled, unlabeled]
   push:
     tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+*' ]
 
 jobs:
   publish:
     if: ${{ github.repository_owner == 'dart-lang' }}
     uses: dart-lang/ecosystem/.github/workflows/publish.yaml@main
+    with:
+      write-comments: false
```
kevmoo, line 17: Don't we WANT comments?

davidmorgan: Hmmmm the build depot has a separate publish comments yaml, so maybe here it does need to write comments. Let's try with just the change above and see what happens :) done.

kevmoo (approved):

PR #791, opened 2025-10-02
Handle null library in _NameTypeChecker.

davidmorgan:
The analyzer docs say "library" can be null mostly on invalid code.

Not sure quite how to trigger it, but I did see this in google3 :)


kevmoo (approved):
I mean, a test would be *nice*...but 🤷

PR #799, opened 2026-01-12
Allow analyzer: '>=8.1.1 <11.0.0'

scheglov:


davidmorgan (approved):
Thanks! Looks good to merge + release.

I did something similar for the build packages just now, tested with dependency override then released.

